PR #25, opened 2021-02-27
Yaml Parser Setup & Tests

### LINKED ISSUE #24: Support for text input parser

mmcgurn:
This is a large PR with code changes including setting up the core code as C based library and a separate c++ client that allows for simple problem setup with YAML parsing. This will close #24


knepley (approved):
I think the refactoring is good. We can see the amount of boilerplate creeping up, which indicates we should be automating physical models at a higher level. It is easy to get too specific and far from the algebra here, so I want us to think for a while first, and just have the courage to rewrite when we get something good. I think we are close, and it will not buy us much until we want to use 5+ models in a run, and maybe do parameter fits.

PR #40, opened 2021-03-26
added error checking and made the boxMesh easier to setup

### LINKED ISSUE #37: Unable to start cases with particles in some configurations
A segfault occurs when particles are not initialized in every domain.

mmcgurn:
Notifies users of the possible error with dm distribution and ranks and makes it explicit for the user to setup.  This closes #37


knepley (approved):
Cool

PR #55, opened 2021-05-21
Support for diffusion in fv compressible code

### LINKED ISSUE #54: Add support for diffusion fluxes into the FVM compressible flow methods

mmcgurn:
Adds support and closes #54


knepley (approved):
This looks like exactly the right idea to me.

I will mention that I see a switch for the aux fields depending on the discretization. You might now be able to use a key to distinguish FV from FEM since aux fields can be retrieved by key.

Issue #66, opened 2021-05-27
Explore 0D chemical reactions in compressible code


mmcgurn:
As part of this work, determine how to include a general Ideal Gas EOS.  Using TChem should be explored with its current PETSc integration. 

- https://www.mcs.anl.gov/petsc/petsc-current/src/ts/tutorials/extchem.c.html
- https://www.mcs.anl.gov/petsc/petsc-current/src/ts/tutorials/extchemfield.c.html

PR #68, opened 2021-05-27
PerfectGas EOS abstraction

### LINKED ISSUE #58: Abstract away concrete equation of state in compressible flows.
The first implementations should include perfect gas

mmcgurn:
Introduces an initial abstraction layer for the EOS used in the compressible formulation and closes #58 .  Note, that the ideal gas implementation of the EOS has been included in issue #66


knepley (approved):
I do not have a good idea whether this is precisely the right interface for EOS, but I think it is clean enough to try out.

mmcgurn:
Thanks I completely agree.  I'm hoping it becomes more clear with the ideal gas from TChem.

PR #71, opened 2021-06-14
ABLATE refactor

### LINKED ISSUE #67: Clean up ablate object model
The flow and particle object models should be cleaned up and unified within the C core library using the new options setup in the eos object

mmcgurn:
This refactor simplifies the code structure in ablateCore/ablateLibrary and unifies how the flow/particles are created.  This PR:
- closes #67 
- starts to address #57

@pakserep Can you take a look at how I changed/updated the particles and tests


ablateLibrary/flow/compressibleFlow.cpp
```diff
@@ -1,34 +1,260 @@
...
+      fluxDifferencer(fluxDifferencerIn == nullptr ? std::make_shared<fluxDifferencer::AusmFluxDifferencer>() : fluxDifferencerIn) {
+    // Create a compressibleFlowData
+    PetscNew(&compressibleFlowData);
+
+    // make sure that the dm works with fv
+    const PetscInt ghostCellDepth = 1;
+    DM& dm = this->dm->GetDomain();
+    {  // Make sure that the flow is setup distributed
+        DM dmDist;
+        DMSetBasicAdjacency(dm, PETSC_TRUE, PETSC_FALSE) >> checkError;
```
knepley, line 50: Is C++ smart enough not to create an object for this stream error check?

ablateLibrary/flow/compressibleFlow.cpp
```diff
@@ -1,34 +1,260 @@
...
+        preStepFunctions.push_back(ComputeTimeStep);
+    }
+
+    // Override the DMTSSetRHSFunctionLocal in DMPlexTSComputeRHSFunctionFVM with a function that includes euler and diffusion source terms
+    DMTSSetRHSFunctionLocal(dm->GetDomain(), CompressibleFlowRHSFunctionLocal, this) >> checkError;
+
+    // copy over any boundary information from the dm, to the aux dm and set the sideset
+    if (auxDM) {
+        PetscDS flowProblem;
+        DMGetDS(dm->GetDomain(), &flowProblem) >> checkError;
```
knepley, line 221: Note that technically a DM can have more than one DS. However, if ABLATE is not going to use that, then we can ignore it.

ablateLibrary/mesh/boxMesh.cpp
```diff
@@ -25,8 +31,8 @@ ablate::mesh::BoxMesh::BoxMesh(std::string name, std::map<std::string, std::stri
         }
     }
 
-    DMPlexCreateBoxMesh(comm, dimensions, simplex ? PETSC_TRUE : PETSC_FALSE, &faces[0], &lower[0], &upper[0], &boundaryTypes[0], PETSC_TRUE, &dm) >> checkError;
-    DMSetOptionsPrefix(dm, name.c_str()) >> checkError;
+    DMPlexCreateBoxMesh(PETSC_COMM_WORLD, dimensions, simplex ? PETSC_TRUE : PETSC_FALSE, &faces[0], &lower[0], &upper[0], &boundaryTypes[0], PETSC_TRUE, &dm) >> checkError;
```
knepley, line 34: Controlling everything explicitly here is probably good. We should consider making another class that just does
```
DMCreate(comm, &dm);
DMSetType(dm, DMPLEX);
DMSetFromOptions(dm);
```
which would allow us to create lots of meshes from PETSc options alone.

knepley (approved):
Looks good. I left a few comments

ablateLibrary/mesh/boxMesh.cpp
mmcgurn: Thanks!  I created a new issue for this #72.

ablateLibrary/flow/compressibleFlow.cpp
mmcgurn: I believe so.  We create a single global instance of the PetscErrorChecker.  The >> operator takes a const reference to the global instance.  I added a constructor/destructor and it appears to only be called once.

PR #87, opened 2021-07-13
Mesh DM and other quality of life improvements 

mmcgurn:
Improvements include:
- default label in flow boundary conditions for FV/FE
- file location including remote url for parsed files 
- input file downloading when url is specified
- dm to read mesh file


knepley (approved):
Looks good. One thing we might consider in the future is allow the YAML to take a URI for a file path. That might make it easier to share consistent setup, since we can just host it on Box.

PR #99, opened 2021-07-22
adds supports for log files into the monitor framework

### LINKED ISSUE #88: Add support for writing to files instead of std::cout
Add support to the monitors to write to a separate log file in the results folder.  This should support writing to CSV

mmcgurn:
this adds support for multiple log types through an interface for the monitors and closes #88


knepley (approved):
I did not completely understand. The documentation is a little weak, but the default PETSc monitoring calls pass a Viewer and format in as the last context argument.

mmcgurn:
> I did not completely understand. The documentation is a little weak, but the default PETSc monitoring calls pass a Viewer and format in as the last context argument.

Thanks!  Is there an example of how this would be done. Is it different call from TSMonitorSet.  The goal for the monitor interface is to allow a user to point to a different output per monitor (l2 norm, time stepping, ignition time) is a reusable declarative way without that monitor instance needed to know output details. I'll work on adding some more documentation on usage before I merge.

knepley:
The default monitors use it: https://www.mcs.anl.gov/petsc/petsc-current/src/snes/interface/snesut.c.html#SNESMonitorDefault

PR #110, opened 2021-08-09
Adds support for setting up/viewering 1D diffusion flames

### LINKED ISSUE #95: Diffusion Flame
Increase testing with support for diffusion flame modeling.

mmcgurn:
this should close #95


mmcgurn:
@knepley I did rebase it, but it turns out to be just a large PR.  The reason it is so big is because I put in a transport model abstraction.

knepley (approved):
It is large, but everything looked good to me. I think encapsulating the transport stuff was a good idea.

I love the CurveMonitor. Maybe that can be hoisted up to PETSc.

PR #122, opened 2021-08-23
L1 norm

sunmJason:
Addition of L1 norm calculation in solutionErrorMonitor


ablateCore/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ add_library(ablateCore "")
 # Include PETSc and MPI
 target_include_directories(ablateCore PUBLIC ${PETSc_INCLUDE_DIRS})
 target_link_libraries(ablateCore PUBLIC ${PETSc_LIBRARIES})
-target_link_directories(ablateCore PUBLIC ${PETSc_LIBRARY_DIRS})
+target_link_directories(ablateCore PUBLIC ${PETSc_LIBRARY_DIRS} /opt/homebrew/Cellar/gcc/11.2.0/lib/gcc/11/)
```
mmcgurn, line 7: Can you remove this line, it is only for m1 Macs.

ablateLibrary/CMakeLists.txt
```diff
@@ -60,7 +60,7 @@ else()
 endif()
 
 ### Ablate Library Setup ###
-add_library(ablateLibrary SHARED "")
+add_library(ablateLibrary SHARED "" flow/boundaryConditions/NeumannGhost.cpp)
```
mmcgurn, line 63: Would you mind reverting this line.  It doesn't look like you are adding any boundary conditions in this PR.

mmcgurn (changes requested):
Thanks for making this PR!  I left a few comments to clean up the code.  You will have to bump the version number in the CMakeLists.txt file.  I would also recommend doing a rebase to reapply this code over the latest in the repo.

I would highly recommend including the l1 norm in some level of testing.  For this I think we could getaway with integration level testing.  Maybe add it to ablate/tests/integrationTests/inputs/compressibleCouetteFlow.yaml and update the expected output in ablate/tests/integrationTests/outputs/compressibleCouetteFlow.txt .  I'd be happy to help with this.

PR #124, opened 2021-08-23
added support for dmViewFromOptions

### LINKED ISSUE #123: Call DMViewFromOptions from the input file
We need to be able to call DMViewFromOptions from the yaml options or when the monitor is used.

mmcgurn:
this adds support for dmViewFromOptions in the input file and closes #123


nathawani (approved):
This is great. Now we can view DM. This is very useful for debugging.

PR #125, opened 2021-08-24
added flow petscClass and fv events

mmcgurn:


knepley (approved):
This looks good. You need to run with

  -log_view :fv.csv:ascii_csv

to get the logfile fv.csv that we parse.

PR #126, opened 2021-08-29
cleaned up the stage/event logging

mmcgurn:
I think I had confused the petscEvent/stage logging.  I have replaced this with a single event for the entire tsSolve that wraps the fv solve.


knepley (approved):
This looks fine

PR #127, opened 2021-08-30
64-bit-index support in ablate

mmcgurn:


ablateLibrary/flow/fvFlow.cpp
```diff
@@ -274,7 +274,7 @@ void ablate::flow::FVFlow::ComputeTimeStep(TS ts, ablate::flow::Flow& flow) {
     }
 
     // take the min across all ranks
-    PetscInt rank;
+    int rank;
```
dsalac, line 277: We should probably use PetscMPIInt, to be consistent.

mmcgurn: Thanks!

ablateLibrary/mathFunctions/parsedSeries.cpp
```diff
@@ -122,7 +122,7 @@ PetscErrorCode ablate::mathFunctions::ParsedSeries::ParsedPetscSeries(PetscInt d
         parser->coordinate[1] = 0;
         parser->coordinate[2] = 0;
 
-        for (auto d = 0; d < std::min(dim, 3); d++) {
+        for (auto d = 0; d < PetscMin(dim, 3); d++) {
```
dsalac, line 125: Suggest changing to
```suggestion
        for (PetscInt d = 0; d < PetscMin(dim, 3); d++) {
```
Similar change on line 130 of ablateLibrary/mathFunctions/parsedNested.cpp

dsalac (approved):
Looks good.

PR #147, opened 2021-10-18
Solver Refactor

mmcgurn:
This refactor is designed to prepare for solvers only using part of the domain. It:

- changes the flow so that a single dmPlex will be used by multiple solvers
- creates a subDomain based upon the plex label
- create a new unified solver interface
- moves the particles to the solver interface (same as flow solvers)

### DIFF
--- a/ablateLibrary/domain/region.hpp
+++ b/ablateLibrary/domain/region.hpp
@@ -1,6 +1,7 @@
 #ifndef ABLATELIBRARY_REGION_HPP
 #define ABLATELIBRARY_REGION_HPP
 
+#include <memory>
 #include <string>
 #include <vector>
 


ablateLibrary/domain/subDomain.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+namespace ablate::domain {
+
+class SubDomain {
+   private:
+    // const pointer to the parent domain/dm
+    const std::weak_ptr<Domain> domain;
+
+    // The label used to describe this subDomain
+    DMLabel label;
```
knepley, line 19: I think you might want (label, value) or even (label, {values}) to specify the subdomain

knepley (changes requested):
Check out my comment about using the label. Otherwise it is great I think.

ablateLibrary/domain/subDomain.hpp
knepley: Yep, that really needs to be fixed. I guess I should fix it now, before anyone else builds code on it.

mmcgurn: @knepley Thanks!  I've added a new struct for a region, that the solvers specify.  This includes a label name and values.  The subDomain now uses a label and values and defaults to "1".

PR #151, opened 2021-11-02
Adds support for subDomain output

mmcgurn:
This PR adds support for subDomain io and arbitrary labeling using a math function.  This moves us closer to #139 and #46

### DIFF
--- a/ablateLibrary/domain/modifiers/distributeWithGhostCells.cpp
+++ b/ablateLibrary/domain/modifiers/distributeWithGhostCells.cpp
@@ -1,0 +1,19 @@
+#include "distributeWithGhostCells.hpp"
+#include <utilities/petscError.hpp>
+
+ablate::domain::modifier::DistributeWithGhostCells::DistributeWithGhostCells(int ghostCellDepthIn) : ghostCellDepth(ghostCellDepthIn < 1 ? 1 : ghostCellDepthIn) {}
+void ablate::domain::modifier::DistributeWithGhostCells::Modify(DM &dm) {
+    // Make sure that the flow is setup distributed
+    DM dmDist;
+    // create any ghost cells that are needed
+    DMSetBasicAdjacency(dm, PETSC_TRUE, PETSC_FALSE) >> checkError;
+    DMPlexDistribute(dm, ghostCellDepth, NULL, &dmDist) >> checkError;
+    if (dmDist) {
+        DMDestroy(&dm) >> checkError;
+        dm = dmDist;
+    }
+}
+
+#include "parser/registrar.hpp"
+REGISTER(ablate::domain::modifier::Modifier, ablate::domain::modifier::DistributeWithGhostCells, "Distribute DMPlex with ghost cells",
+         OPT(int, "ghostCellDepth", "the number of ghost cells to share on the boundary.  Default is 1."));
--- a/ablateLibrary/domain/subDomain.cpp
+++ b/ablateLibrary/domain/subDomain.cpp
@@ -1,7 +1,8 @@
 #include "subDomain.hpp"
 #include <utilities/petscError.hpp>
 
-ablate::domain::SubDomain::SubDomain(std::weak_ptr<Domain> domain, std::shared_ptr<domain::Region> region) : domain(domain), region(region), name(""), label(nullptr), auxDM(nullptr), auxVec(nullptr) {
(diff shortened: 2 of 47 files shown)


ablateLibrary/domain/modifiers/distributeWithGhostCells.cpp
```diff
@@ -0,0 +1,19 @@
+#include "distributeWithGhostCells.hpp"
+#include <utilities/petscError.hpp>
+
+ablate::domain::modifier::DistributeWithGhostCells::DistributeWithGhostCells(int ghostCellDepthIn) : ghostCellDepth(ghostCellDepthIn < 1 ? 1 : ghostCellDepthIn) {}
+void ablate::domain::modifier::DistributeWithGhostCells::Modify(DM &dm) {
+    // Make sure that the flow is setup distributed
+    DM dmDist;
+    // create any ghost cells that are needed
+    DMSetBasicAdjacency(dm, PETSC_TRUE, PETSC_FALSE) >> checkError;
+    DMPlexDistribute(dm, ghostCellDepth, NULL, &dmDist) >> checkError;
```
knepley, line 10: Note that this is one option for having overlap, but another is to just add the call to `DMPlexDistributeOverlap()` after a regular distribution.

ablateLibrary/domain/subDomain.cpp
```diff
@@ -153,3 +220,158 @@ void ablate::domain::SubDomain::ProjectFieldFunctions(const std::vector<std::sha
...
+
+    // March over the global section
+    PetscSection section;
+    DMGetGlobalSection(sDM, &section) >> checkError;
+
+    PetscInt pStart, pEnd;
+    PetscSectionGetChart(section, &pStart, &pEnd) >> checkError;
+
+    // For each field in the subDM
+    for (std::size_t i = 0; i < subFields.size(); i++) {
```
knepley, line 343: Note that this mostly reproduces the code in `DMCreateSubDM()` which returns an IS with these indices.

ablateLibrary/domain/subDomain.cpp
```diff
@@ -137,7 +166,45 @@ PetscObject ablate::domain::SubDomain::GetPetscFieldObject(const Field& field) {
...
+
+    if (label == nullptr) {
+        DMProjectFunction(dm, time, &fieldFunctions[0], &fieldContexts[0], INSERT_VALUES, globVec) >> checkError;
+
+    } else {
+        DMProjectFunctionLabel(dm, time, label, region->GetValues().size(), &region->GetValues()[0], -1, NULL, &fieldFunctions[0], &fieldContexts[0], INSERT_VALUES, globVec);
+    }
+}
+
+void ablate::domain::SubDomain::ProjectFieldFunctionsToSubDM(const std::vector<std::shared_ptr<mathFunctions::FieldFunction>>& initialization, Vec globVec, PetscReal time) {
```
knepley, line 196: Note that the `dm` input to Project can have different fields than the DM attached to the input vector, so you can compress out fields with the single call if you want.

knepley (approved):
I made some comments

ablateLibrary/domain/subDomain.cpp
mmcgurn: Thanks!  Does that mean that the fieldFunctions should be sized/correspond to vec and not the dm?

ablateLibrary/domain/subDomain.cpp
mmcgurn: Thanks @knepley!  Are there any support functions for moving data with the IS or should I iterate through it accessing array data?

mmcgurn:
> I made some comments

Thanks!  I think I'm going to merge this in for now and go back and cleanup and use DMCreateSubDM in the future.

ablateLibrary/domain/subDomain.cpp
knepley: Yes, you call [VecISCopy](https://petsc.org/main/docs/manualpages/Vec/VecISCopy.html), (VecISSet)[https://petsc.org/main/docs/manualpages/Vec/VecISSet.html], [VecISAXPY](https://petsc.org/main/docs/manualpages/Vec/VecISAXPY.html)

ablateLibrary/domain/subDomain.cpp
knepley: I do not completely understand the question, but we could go over it at the meeting.

knepley:
Sounds good.

PR #152, opened 2021-11-07
Changed field definition to a pre-solver step

mmcgurn:
This PR changed the field specification to a pre solver step allowing us to use the same fields in different solvers.


ablateLibrary/domain/fieldDescription.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+PetscObject ablate::domain::FieldDescription::CreatePetscField(DM dm) const {
+    switch (type) {
+        case FieldType::FEM: {
+            // determine if it a simplex element and the number of dimensions
+            DMPolytopeType ct;
+            PetscInt cStart;
+            DMPlexGetHeightStratum(dm, 0, &cStart, NULL) >> checkError;
+            DMPlexGetCellType(dm, cStart, &ct) >> checkError;
+            PetscInt simplex = DMPolytopeTypeGetNumVertices(ct) == DMPolytopeTypeGetDim(ct) + 1 ? PETSC_TRUE : PETSC_FALSE;
```
knepley, line 30: You can just use DMPlexIsSimplex() here I think.

ablateLibrary/domain/fileMesh.cpp
```diff
@@ -1,30 +1,24 @@
 #include "fileMesh.hpp"
 #include <utilities/petscError.hpp>
 #include <utilities/petscOptions.hpp>
-ablate::domain::FileMesh::FileMesh(std::string nameIn, std::filesystem::path pathIn, std::shared_ptr<parameters::Parameters> options, std::vector<std::shared_ptr<modifier::Modifier>> modifiers)
```
knepley, line 4: Note that you can do this using `-dm_plex_filename`, so it is unclear whether we need this class, but maybe due to how the YAML is processed.

knepley (approved):
I tried to go over everything. I think it looks good. I made a few comments.

ablateLibrary/domain/fileMesh.cpp
mmcgurn: Thanks!  It is not really needed, it must simplifies the setup in yaml so yaml is self explaining for the examples.

ablateLibrary/domain/fieldDescription.cpp
mmcgurn: Thanks!  I updated the code.

Issue #155, opened 2021-11-10
Add support for Restart with SubSolvers


mmcgurn:
This is a follow on to issue #139

PR #156, opened 2021-11-10
Support for subSolver with FVM

mmcgurn:
this pr adds support for subSolver usage with the FVM and adds a simple test closing #139.  I did separate out part of this work to #155.


knepley (approved):
This looks right to me. I wonder if we should move some of the label manipulation into the library.

PR #161, opened 2021-11-17
Boundary Solver

mmcgurn:
Adds a new boundary solver (no physics yet) but moves us closer to the one-dimensional inviscid boundary conditions to Euler FV #46

### DIFF
--- a/ablateLibrary/boundarySolver/boundarySolver.cpp
+++ b/ablateLibrary/boundarySolver/boundarySolver.cpp
@@ -1,0 +1,314 @@
+#include "boundarySolver.hpp"
+#include <set>
+#include <utility>
+#include "boundaryProcess.hpp"
+
+ablate::boundarySolver::BoundarySolver::BoundarySolver(std::string solverId, std::shared_ptr<domain::Region> region, std::shared_ptr<domain::Region> fieldBoundary,
+                                                       std::vector<std::shared_ptr<BoundaryProcess>> boundaryProcesses, std::shared_ptr<parameters::Parameters> options)
+    : Solver(std::move(solverId), std::move(region), std::move(options)), fieldBoundary(std::move(fieldBoundary)), boundaryProcesses(std::move(boundaryProcesses)) {}
+
+ablate::boundarySolver::BoundarySolver::~BoundarySolver() {
+    if (gradientCalculator) {
+        PetscFVDestroy(&gradientCalculator);
+    }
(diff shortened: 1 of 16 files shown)


ablateLibrary/boundarySolver/boundarySolver.cpp
```diff
@@ -13,6 +13,47 @@ ablate::boundarySolver::BoundarySolver::~BoundarySolver() {
...
+
+    // Check to see if this cell is already in the list
+    if (stencilSet.count(cell)) {
+        return;
+    }
+
+    // Add to the list
+    stencilSet.insert(cell);
+
+    // If not at max depth, check another layer
```
knepley, line 27: This is a perfectly fine way of doing it. I will just note that we have https://petsc.org/main/docs/manualpages/DMPLEX/DMPlexGetAdjacency.html ,which works in a similar way.

ablateLibrary/boundarySolver/boundarySolver.cpp
```diff
@@ -312,3 +321,63 @@ PetscErrorCode ablate::boundarySolver::BoundarySolver::ComputeRHSFunction(PetscR
     VecRestoreArrayRead(cellGeomVec, &cellGeomArray) >> checkError;
     PetscFunctionReturn(0);
 }
+void ablate::boundarySolver::BoundarySolver::InsertFieldFunctions(const std::vector<std::shared_ptr<mathFunctions::FieldFunction>>& fieldFunctions, PetscReal time) {
```
knepley, line 365: This also seems fine, but you might be able to do the same thing with DMProjectFunction(). If you can't, I would be interested, because it means I might be able to extend it in some way.

knepley (approved):
I made some comments

ablateLibrary/boundarySolver/boundarySolver.cpp
mmcgurn: I'm projecting to the cell face.  For cells with multiple faces I'm currently representing the face as the average centroid location and the sum of the normal areas.

ablateLibrary/boundarySolver/boundarySolver.cpp
mmcgurn: Thanks!  I went the more manual up/down/up approach so that I could check the face and only add cells that are on the other side of boundary faces.

PR #164, opened 2021-12-01
update for changes in petsc dmplex api

mmcgurn:


knepley (approved):
I did not love the output update, but Barry says it is clarer

PR #189, opened 2022-01-19
Maesemen/riemann stiff

### LINKED ISSUE #186: Riemann Solver for Stiffened equation of state
Further modify the existing Riemann solver for stiffened gas

maesemen:
Adds stiffened gas Riemann solver, and edited EOS to match Change and Liou 2007 model. Closes #186 
Note: only one unit test for riemannStiff using stiff EOS, might need to add more to generalize and check initial guess.


jikembo (approved):

mmcgurn (commented):
Everything looks good.  I think that you maybe able to simplify the use of the code by having a single Riemann flux calculator and just return the correct function for either a stiff/perfect gas from GetFluxCalculatorFunction() based upon the type of eos passed in.  This would allow a single class to cover all cases.

PR #191, opened 2022-01-20
set the package config path directly in env for clion

mmcgurn:
This removes the influence of local env variables and sets the package config path correctly when using presets in clion.


jikembo (commented):
Have you done this before? Is this a new change?

jikembo (approved):

mmcgurn:
> Have you done this before? Is this a new change?

The  package config path was not correctly before.

PR #194, opened 2022-01-22
Maesemen/stiff decode

### LINKED ISSUE #188: Decoding for Stiffened Gas Equation of State
Decoding state of equation for stiffened gas

maesemen:
fixes bug in twoPhaseAdvection. adds analytical decode state for perfect gas + stiffened gas case. adds integration test for 2D Sod problem.  closes #188


jikembo (commented):
In the quadratic formula for roots in the decode state, PESTcmax was used for positive roots. However, petsmax is only for max of the two roots. What if the roots are negative or can we be sure both roots are always positive! You may want to have some caution there.

ablateLibrary/eos/stiffenedGas.hpp
```diff
@@ -54,6 +54,7 @@ class StiffenedGas : public EOS {
     ComputeSpecificHeatFunction GetComputeSpecificHeatConstantVolumeFunction() override { return StiffenedGasComputeSpecificHeatConstantVolume; }
     void* GetComputeSpecificHeatConstantVolumeContext() override { return &parameters; }
     PetscReal GetSpecificHeatRatio() const { return parameters.gamma; }
+    PetscReal GetSpecificHeatCp() const { return parameters.Cp; }
```
mmcgurn, line 57: There is already a GetComputeSpecificHeatConstantVolumeFunction and GetComputeSpecificHeatConstantVolumeContext to get the specific heat.  We may want to just use these interface functions to help future proof any calls.

tests/integrationTests/inputs/shockTube2Gas2D.yaml
```diff
@@ -0,0 +1,90 @@
+---
+environment:
+  title: SOD_2Gas_6x6_ArAir
+  tagDirectory: true
+arguments: {}
+timestepper:
+  name: theMainTimeStepper
+  arguments:
+    ts_type: rk
+    ts_max_time: 0.01
```
mmcgurn, line 10: You dt is of fixed size and adapt is off.  This may be ok, but you have to make sure you don't step over the cfl constraint.

mmcgurn (changes requested):
Look good!  I made one small change request.

mmcgurn (approved):

jikembo (approved):

PR #215, opened 2022-02-03
Maesemen/riemann stiff pstar

### LINKED ISSUE #213: RiemannStiff convergence issue and stiff decode state add error catch
issue with pstar not converging with RiemannStiff flux calculator
stiff decode not catching all possible zero values

maesemen:
Fixes pstar convergence. closes #213

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@ if(NOT DEFINED CMAKE_CXX_COMPILER)
 endif()
 
 # Set the project details
-project(ablateLibrary VERSION 0.7.32)
+project(ablateLibrary VERSION 0.7.33)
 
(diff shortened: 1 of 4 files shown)


jikembo (approved):

mmcgurn (changes requested):
For bugs like this, I think that it is a good idea to add a test case that would have caught this error.  Is there a test case you can add to catch this?

maesemen:
> For bugs like this, I think that it is a good idea to add a test case that would have caught this error. Is there a test case you can add to catch this?

It would take some work to figure out the conditions for the test case to catch this, but I did check with running a full simulation

jikembo (approved):

PR #223, opened 2022-02-16
Preliminary support for chemTab Chemistry Model

mmcgurn:
@chandola I've rebased your branch and I think I got the dependencies working.  Some of the tests were not passing for me, I think that it was because zmix is now considered part of the progress variable list.  I've added these zMix to the progress variable list and most of the tests are now passing, but I'm not sure how to update ChemTabModelComputeProgressVariablesFixture.  Any help fixing this test would be appreciated.


tests/ablateLibrary/inputs/chemistry/chemTabTestModel_1/weights.csv
```diff
@@ -0,0 +1,54 @@
+,zmix,CPV_0,CPV_1
```
chandola, line 1: The new ChemTab weights file will not include zmix as a column. This is because we do not project the mass fraction vector to get zmix. I have pushed in the updated weights file as well as the inverse weights file.

chandola (commented):
The weights file in the ChemTab model (weights.csv) needs to be updated. Latest version is pushed in. Also included is the weights_inv.csv file that contains the pseudo-inverse of the weights matrix, used to calculate mass fractions from the progress variables.

chandola (approved):

PR #242, opened 2022-03-09
Add interface to support multiple probe init techniques

### LINKED ISSUE #239: Add multiple probe initiation methods
Allow multiple probe initiation methods like:
- list of locations/names
- line/rake
- etc

mmcgurn:
this PR
- adds support for probe rakes and closes #239
- removes build directions from doc and moves to wiki.  This is so that the directions can be updated faster (without PR) and more easily contributed to by multiple people.


knepley (approved):
Cool. Sometime we might want to combine the grid-like thing here with the Monte-Carlo particle initialization we just put into PETSc.

PR #249, opened 2022-04-06
eos refactor and cleanup

### LINKED ISSUE #230: Unify aux field information/update for FV solutions
Simplify/update aux field information so that energy, pressure, velocity, yi, viscosity, diffusion, etc is always assumed store in an aux field and does not need to be decoded,

mmcgurn:
This is a major refactor of the EOS. The equation of state is now designed to to compute thermodynamic properties based upon the conserved field variables being solved.  This can range from euler & densityYi and in the future  euler & progresses variables with ChemTab.  This addresses and closes #230 . @maesemen Can you take a look at the two phase changes please and think about how we could create a twoPhase eos?


knepley (approved):
I could not review everything, but it looks good to me. It took me a minute to see that you are defining lambdas and passing them back.

I am interested to understand the two-phase advection test

mmcgurn:
> I could not review everything, but it looks good to me. It took me a minute to see that you are defining lambdas and passing them back.
> 
> I am interested to understand the two-phase advection test

I find using lambdas really helpful for testing and initialization.

PR #250, opened 2022-04-07
adding dakota integration document

danial-faghihi:
The Dakota documentation is not included in the list of the contents.


danial-faghihi:
> @danial-faghihi Thank you for this! Would you mind making a few small changes. We need to bump the version number in the main CMakeLists.txt. Could you add a header to the markdown file that tells the web builder how to layout it. Something like the following (with the ---).
> 
> ```yaml
> ---
> layout: default
> title: DAKOTA Integration
> parent: Development Guides
> nav_order: 13
> ---
> ```

PR #252, opened 2022-04-11
Corrected two phase fluxes

maesemen:
Corrects the upwinding for the stratified flow model, using the upwind direction for each section of the interface instead of one for the whole face.  Also uses the interface pressure instead of upwind pressure for more accurate calculations.  Adds printout of total runtime of code.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -1,6 +1,6 @@
+#include <time.h>
 #include <fstream>
 #include <iostream>
-#include <time.h>
 #include <memory>
 #include <parameters/petscPrefixOptions.hpp>
 #include <utilities/fileUtility.hpp>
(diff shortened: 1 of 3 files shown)


main.cpp
```diff
@@ -103,4 +107,9 @@ int main(int argc, char** args) {
         }
     }
     PetscFinalize() >> checkError;
+
+    // output runtime
+    clock_t stop = clock();
+    double elapsed = (double)(stop - start) / CLOCKS_PER_SEC;
+    std::cout << "Runtime: " << elapsed << " sec" << std::endl;
```
mmcgurn, line 114: @maesemen I would prefer if we needed add any required output that is not controlled by the user.  If we wanted to do this I think that we should allow them to specify if/how the output is created.  However, this functionality is already covered in PETSc with more detail.  Have you used  log_view https://petsc.org/main/docs/manualpages/Profiling/PetscLogView.html ?

mmcgurn (changes requested):

main.cpp
maesemen: I took it out for now.  I have not used log_view but I will take a look.

mmcgurn (approved):

mmcgurn:
@maesemen Thanks!  It looks like the test issue has been resolved.

PR #265, opened 2022-05-05
multiple grad interpolant support

### LINKED ISSUE #245: Can save/restore PGS vector with 1200 nodes on Quartz

mmcgurn:
This PR adds support for multiple interpolants with the same stencil in the FVM model.  Changes may fix and close #245.


knepley (approved):
Okay,  I think I see. You make a general function that packs up the field values on some stencil and then you feed it to functions that consume the stencil, like a least-square fit.

PR #281, opened 2022-06-03
bug fix for rad solver

mmcgurn:
@owenmylotte I think I found a bug in the rad solver tests.  On line 62 we were reaching past the end of the array.  I've changed this to the latest item.  Is this correct?

### DIFF
--- a/tests/ablateLibrary/radiation/radiationTests.cpp
+++ b/tests/ablateLibrary/radiation/radiationTests.cpp
@@ -10,6 +10,7 @@
 #include "convergenceTester.hpp"
 #include "domain/boxMesh.hpp"
 #include "domain/modifiers/ghostBoundaryCells.hpp"
+#include "domain/modifiers/setFromOptions.hpp"
 #include "eos/perfectGas.hpp"
 #include "eos/radiationProperties/constant.hpp"
 #include "eos/radiationProperties/radiationProperties.hpp"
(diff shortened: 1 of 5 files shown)


tests/ablateLibrary/radiation/radiationTests.cpp
```diff
@@ -58,7 +59,7 @@ static PetscReal CSimp(PetscReal a, PetscReal b, std::vector<double> f) {
             }
             f_sum += f[i];  //!< Add this value to the total every time
         }
-        I = ((b - a) / (3 * n)) * (f[0] + f_sum + f[n]);  //!< Compute the total final integral
+        I = ((b - a) / (3 * n)) * (f[0] + f_sum + f[n - 1]);  //!< Compute the total final integral
```
owenmylotte, line 62: This would cause a problem, thank you for catching that.

tests/ablateLibrary/radiation/radiationTests.cpp
```diff
@@ -178,7 +179,8 @@ TEST_P(RadiationTestFixture, ShouldComputeCorrectSourceTerm) {
 
         auto domain = std::make_shared<ablate::domain::BoxMesh>("simpleMesh",
                                                                 fieldDescriptors,
-                                                                std::vector<std::shared_ptr<ablate::domain::modifiers::Modifier>>{},
+                                                                std::vector<std::shared_ptr<ablate::domain::modifiers::Modifier>>{std::make_shared<ablate::domain::modifiers::SetFromOptions>(
```
owenmylotte, line 182: Thank you for these other fixes also!

owenmylotte (approved):
That is correct, thank you! I hadn’t realized the size issue while writing that and I’m surprised I didn’t notice issues. Thanks again for your additions.

PR #283, opened 2022-06-07
added support for downloading files/folders from github

mmcgurn:
this refactor adds support for downloading files/folders from GitHub and allows this to be expanded to different resources.  This implementation bypasses the need for an http/https client and uses the existing curl command found in petsc.


dsalac (approved):
Looks good. I noticed that https:// is there twice in download.hpp, line 19.

knepley (commented):
Looks good to me

PR #288, opened 2022-06-21
Chemtab new

profPlum:
Hey Matt, this is an update to the chemtab branch that contains one of the latest models
& new code to accommodate testing/integrating it.

I had thought this code was already pushed & integrated... Please use this rather than what is there already.


mmcgurn:
Hi @profPlum.  It looks like there are conflicts that must be resolved.  Can you resolve these conflicts with a rebase before I review them.  It looks like some of these commits have already been merged from this branch, you may want to rebase with only commits after 3c7c1a0.

profPlum:
@mmcgurn I believe I resolved the conflicts

mmcgurn:
@profPlum Do you see this error? 
image

profPlum:
@mmcgurn should be ready for merge now

tests/ablateLibrary/chemistry/adapt_test_targets.py
```diff
@@ -0,0 +1,62 @@
+import numpy as np
```
mmcgurn, line 1: How is this python being used with ablate.  Should it instead live in the chemTabRepo?

mmcgurn (changes requested):

tests/ablateLibrary/chemistry/adapt_test_targets.py
profPlum: It belongs here, it automatically adapts the tests for the model to a new model.
It dynamically writes a c++ header file.

mmcgurn: So it is used to generate the test files.  Lets move this out of this repo and keep it with chemTab where we assume python is available.

tests/ablateLibrary/chemistry/chemTabModelTests.cpp
```diff
@@ -70,12 +73,7 @@ TEST_P(ChemTabModelGetSpeciesAndProgressVariableTestFixture, ShouldReturnCorrect
...
-                             .modelPath = "inputs/chemistry/chemTabTestModel_1",
-                             .expectedSpecies = {"YiCH4",   "YiH",     "YiO",    "YiO2",   "YiOH",    "YiH2O",  "YiHO2",   "YiH2O2",   "YiC",      "YiCH",   "YiCH2",  "YiCH2(S)", "YiCH3",  "YiH2",
-                                                 "YiCO",    "YiCO2",   "YiHCO",  "YiCH2O", "YiCH2OH", "YiCH3O", "YiCH3OH", "YiC2H",    "YiC2H2",   "YiC2H3", "YiC2H4", "YiC2H5",   "YiC2H6", "YiHCCO",
-                                                 "YiCH2CO", "YiHCCOH", "YiN",    "YiNH",   "YiNH2",   "YiNH3",  "YiNNH",   "YiNO",     "YiNO2",    "YiN2O",  "YiHNO",  "YiCN",     "YiHCN",  "YiH2CN",
-                                                 "YiHCNN",  "YiHCNO",  "YiHOCN", "YiHNCO", "YiNCO",   "YiC3H7", "YiC3H8",  "YiCH2CHO", "YiCH3CHO", "YiN2",   "YiAR"},
-                             .expectedProgressVariables = {"zmix", "CPV_0", "CPV_1", "CPV_2", "CPV_3"}}));
+                             .modelPath = "inputs/chemistry/chemTabTestModel_1", .expectedSpecies = {SPECIES_NAMES}, .expectedProgressVariables = {CPV_NAMES}}));
```
mmcgurn, line 76: It looks like these tests are setup to run multiple themTabTestModels so hardcoding having a single set of SPECIES_NAMES,  CPV_NAMES, etc won't allow for testing multiple models. If we want to have a file containing test results I think a text file like yaml (which can still be auto generated)that is parsed by the test code might work better.  There is already a yaml parser in the code.  That way we could store it in the model directory but pass the path in the test params like the model.

mmcgurn (changes requested):

tests/ablateLibrary/chemistry/chemTabModelTests.cpp
profPlum: Yes that is why I created the adapt_model_tests.py script, it generates a header file with macros for the test targets. The header file is then included inside the cpp test file & used for comparison.

mmcgurn: It doesn't work very well as automated testing parameters. These are automated test that run in ablate, not something that you modify each time.  ChemTabModelTests is parameterized so that they can run multiple sets of input/outputs without recompiling or regenerating the tests.  This can't be done with a single header file with non-specific defines. In the future we may also want to use an ablate library executable to test the models in chemTabRepo we are going to populate, we can't do this with compile time constants.  In order for this to be implemented in ablate, please change the tests to allow for multiple parameter to be tested.  If you want to use autogenerated test inputs/outputs please use runtime parameters such as an yaml file.

profPlum: You don't need to worry about how the header file is generated.
What I'm doing is _exactly what Varun's already merged code does_ except now: it is more organized,
a bunch of related constants are grouped together in one header file (rather than baked inline) &
it has a much better model than the previous one.

I think we should prioritize actually testing the model.
But if _you_ really want to add yaml files, you can feel free.

mmcgurn (approved):

CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ if (NOT DEFINED CMAKE_CXX_COMPILER)
 endif ()
 
 # Set the project details
-project(ablateLibrary VERSION 0.8.28)
+project(ablateLibrary VERSION 0.8.38)
```
mmcgurn, line 19: We need to increment the versions.

mmcgurn (changes requested):

CMakeLists.txt
mmcgurn: it should be 29

mmcgurn (approved):

PR #294, opened 2022-07-06
PETSc Particle Radiation

owenmylotte:
Allows for parallel implementation of the radiation solver.


tests/integrationTests/outputs/radiation/parallelPlatesOutput.txt
```diff
@@ -2,18 +2,360 @@ MinMaxAvg euler for timestep 0000:
 	min: [0.177, 2.53e+05,  0,  0]
 	max: [0.623, 2.53e+05,  0,  0]
 	avg: [0.244, 2.53e+05,  0,  0]
+DMSwarmDataEx: instance=0
```
mmcgurn, line 5: What controls the dmswarm data output?

src/radiation/radiation.cpp
```diff
@@ -204,92 +451,285 @@ PetscErrorCode ablate::radiation::Radiation::ComputeRHSFunction(PetscReal time,
...
+        EndEvent();
+        VecRestoreArrayRead(cellGeomVec, &cellGeomArray) >> checkError;
+    }
+
     PetscFunctionReturn(0);
 }
 
-PetscReal ablate::radiation::Radiation::FlameIntensity(double epsilon, double temperature) { /** Gets the flame intensity based on temperature and emissivity*/
+PetscReal ablate::radiation::Radiation::FlameIntensity(double epsilon, double temperature) { /** Gets the flame intensity based on temperature and emissivity (black body intensity) */
     const PetscReal sbc = 5.6696e-8;                                                         //!< Stefan-Boltzman Constant (J/K)
```
mmcgurn, line 708: Lets move these contacts to a static class in the utilities folder.

src/radiation/radiation.cpp
```diff
@@ -204,92 +451,285 @@ PetscErrorCode ablate::radiation::Radiation::ComputeRHSFunction(PetscReal time,
...
     const PetscReal pi = 3.1415926535897932384626433832795028841971693993;
```
mmcgurn, line 709: It looks like we have multiple definitions of Pi and some other constants, can we merge them into a single state class in utilities.

src/radiation/radiation.cpp
```diff
@@ -204,92 +451,285 @@ PetscErrorCode ablate::radiation::Radiation::ComputeRHSFunction(PetscReal time,
...
     return epsilon * sbc * temperature * temperature * temperature * temperature / pi;
 }
 
+std::string ablate::radiation::Radiation::Key(Identifier id) {  //!< Nested Cantor pairing function in order to identify ray segment
```
mmcgurn, line 713: You may want to change Identifier to const Identifier& to prevent memory copies.  We may want to explore changing the id from a string to something else in the future.

mmcgurn (changes requested):
Looks great!  I've made some recommendations.

src/radiation/radiation.cpp
```diff
@@ -34,166 +89,358 @@ void ablate::radiation::Radiation::Initialize() {
...
+             * Update the coordinates of the particle (virtual and physical)
+             * */
+            virtualcoord[ipart].x += virtualcoord[ipart].xdir;  //!< x component: add one step to the coordinate position
+            virtualcoord[ipart].y += virtualcoord[ipart].ydir;  //!< y component: add one step to the coordinate position
+            virtualcoord[ipart].z += virtualcoord[ipart].zdir;  //!< z component: add one step to the coordinate position
+
+            UpdateCoordinates(ipart, virtualcoord, coord);  //!< Update the particle coordinates into the physical coordinate system
+        }
+        /** Restore the fields associated with the particles after all of the particles have been stepped */
+        DMSwarmRestoreField(radsearch, DMSwarmPICField_coor, NULL, NULL, (void**)&coord) >> checkError;
```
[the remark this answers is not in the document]
owenmylotte: This is how I've seen it used in the PETSc examples and I don't believe there are any other candidates (it doesn't compile without), but I'm not sure why the (void**) is necessary.

src/radiation/radiation.cpp
owenmylotte: Creating a struct equality operator for the ray identities is on my to do list. I can do that while face stepping stuff is coming down the chain.

mmcgurn: Sounds good!  I don't think that we need to hold up the PR for this.

src/utilities/constants.hpp
```diff
@@ -0,0 +1,10 @@
+#ifndef ABLATELIBRARY_CONSTANTS_HPP
+#define ABLATELIBRARY_CONSTANTS_HPP
+
+#include <petscsystypes.h>
+
+/// Class Constants
```
mmcgurn, line 6: Can we wrap this in a class to prevent name conflicts?  Something like:

`namespace ablate::utilities {
class VectorUtilities {
   public:
     static const PetscReal sbc = 5.6696e-8;  //!< Stefan-Boltzman Constant (J/K)
 static const PetscReal pi = 3.1415926535897932384626433832795028841971693993;

mmcgurn (changes requested):

mmcgurn (approved):

PR #317, opened 2022-08-10
Radiation Abstraction

owenmylotte:
Abstracts the radiation machinery to allow for more general use than finite volume energy. (Such as surface flux implementation for regression rate)


CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ if (CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
 endif ()
 
 # Create the library for ablate
-add_library(ablateLibrary SHARED)
+add_library(ablateLibrary SHARED src/radiation/volumeRadiation.cpp src/radiation/volumeRadiation.hpp)
```
mmcgurn, line 41: Would you mind moving these to the the cmakelist in the radiation folder?

src/radiation/radiation.cpp
```diff
@@ -409,16 +331,11 @@ void ablate::radiation::Radiation::RayInit() {
...
-    RestoreRange(cellRange);
     DMDestroy(&radsearch) >> checkError;
 
     if (log) EndEvent();
 }
-
-PetscErrorCode ablate::radiation::Radiation::ComputeRHSFunction(PetscReal time, Vec solVec, Vec rhs) {
-    PetscFunctionBeginUser;
-
+const std::map<PetscInt, ablate::radiation::Radiation::Origin>& ablate::radiation::Radiation::Solve(Vec solVec, Vec rhs) {
```
mmcgurn, line 338: What does the radiation solver do with the solveVec and rhs?

owenmylotte: Hi, the absorptivity function needs the solution vector for mass fraction information. However, the solve no longer needs the rhs vector and I've removed this now.

mmcgurn (approved):

PR #320, opened 2022-08-12
Gas Absorption Model Input Fix

owenmylotte:
Fix string comparison issue for species mass fractions in input file. (Upper / lower case sensitivity)


mmcgurn (changes requested):
Looks good.  For fixes like this I would recommend putting in test cases in radiationZimmerTests.cpp that would have failed before this fix.  It helps prevent this bug from creeping back in.

mmcgurn (approved):

PR #323, opened 2022-08-15
Face Stepping Initialization

owenmylotte:
Speed and accuracy improvements for the radiation initialization.


src/radiation/radiation.cpp
```diff
@@ -330,8 +374,11 @@ void ablate::radiation::Radiation::Initialize(solver::Range cellRangeIn) {
             stepcount++;
         }
     }
-    /** Cleanup*/
+    /** Cleanup */
     DMDestroy(&radsearch) >> checkError;
+    VecRestoreArrayRead(faceGeomVec, &faceGeomArray);
```
mmcgurn, line 379: Would you mind adding the error checking here?

src/radiation/volumeRadiation.cpp
```diff
@@ -4,7 +4,9 @@
 ablate::radiation::VolumeRadiation::VolumeRadiation(const std::string& solverId1, const std::shared_ptr<domain::Region>& region, std::shared_ptr<domain::Region> fieldBoundary,
                                                     const PetscInt raynumber, const std::shared_ptr<parameters::Parameters>& options,
                                                     std::shared_ptr<eos::radiationProperties::RadiationModel> radiationModelIn, std::shared_ptr<ablate::monitors::logs::Log> log)
-    : Radiation(solverId1, region, fieldBoundary, raynumber, options, radiationModelIn, log), CellSolver(solverId1, region, options) {}
+    : Radiation(solverId1, region, fieldBoundary, raynumber, options, radiationModelIn, log),
+      CellSolver(solverId1, region, options),
+      solverRegionMinusGhost(std::make_shared<domain::Region>(solverId1 + "_minusGhost")) {}
```
mmcgurn, line 9: Is the solverRegionMinusGhost being used anywhere?

mmcgurn (changes requested):
Looks good, just some small questions.

src/radiation/volumeRadiation.cpp
owenmylotte: Ah, I had set the VolumeRadiation up to omit the ghost cells. During debugging I realized it made no difference so I removed it. I can get rid of that label.

mmcgurn (approved):

PR #329, opened 2022-08-23
Radiation MPI Fix

owenmylotte:
Push particles with invalid face geometries. Solve radiation at specified interval only.


src/radiation/radiation.cpp
```diff
@@ -240,9 +237,14 @@ void ablate::radiation::Radiation::Initialize(solver::Range cellRangeIn) {
              * Assemble a vector of vectors etc associated with each cell index, angular coordinate, and space step?
              * The boundary has been reached if any of these conditions don't hold
              * */
+
```
mmcgurn, line 240: Would you mind removing the debug code?

mmcgurn (changes requested):
Sorry for the late review.  Looks good just one request.  Did you check the clang-tidy messages?

Thanks!

src/radiation/radiation.cpp
owenmylotte: Ah, I didn't realize that was still in there. I can do that.

mmcgurn (approved):

mmcgurn (approved):

Issue #284, opened 2022-06-08
Implement DM Refine Hook


mmcgurn:

mmcgurn:
https://petsc.org/main/docs/manualpages/DM/DMRefineHookAdd/

PR #344, opened 2022-09-15
added init call to boundary process

### LINKED ISSUE #342: SubDomain as Shared Pointer
The subDomain must be able to be cast as a shared pointer for moving between classes.

mmcgurn:
@owenmylotte this should address your problem.  This also sets up the boundary to mirror the other procces/solver and will take close #342  and will make #284 easier.


owenmylotte (approved):
It works great thank you!

PR #346, opened 2022-09-19
Fix for the Zimmer absorption model in zero density.

owenmylotte:
Sets the absorption to zero in a vacuum. Prevents nans for intensity from boundary cells where the density is set to zero.


tests/unitTests/eos/radiationProperties/radiationZimmerTests.cpp
```diff
@@ -75,4 +75,10 @@ INSTANTIATE_TEST_SUITE_P(RadationZimmerTests, ZimmerTestFixture,
                                                                 .conservedValues = {0.5, NAN, NAN, NAN, NAN, 0.0015, 0.0005, 0.0035, 0.0045},  //!< The Density Yi values live here
                                                                 .temperatureIn = 1200.0,
                                                                 .densityIn = 1.1,
```
mmcgurn, line 77: Nice job adding a test to catch this!

mmcgurn (approved):

PR #347, opened 2022-09-19
Sublimation Radiation Surface Flux

owenmylotte:
Adds surface radiation capability to the sublimation solver to account for the effect of radiation on regression rate. Adds one dimensional capability to the radiation solver.

### DIFF
--- a/src/boundarySolver/lodi/isothermalWall.cpp
+++ b/src/boundarySolver/lodi/isothermalWall.cpp
@@ -1,7 +1,5 @@
 #include "isothermalWall.hpp"
 #include "finiteVolume/compressibleFlowFields.hpp"
-#include "mathFunctions/functionFactory.hpp"
-#include "solver/dynamicRange.hpp"
 #include "utilities/mathUtilities.hpp"
 
 using fp = ablate::finiteVolume::CompressibleFlowFields;
--- a/src/boundarySolver/lodi/isothermalWall.cpp
+++ b/src/boundarySolver/lodi/isothermalWall.cpp
@@ -1,10 +1,13 @@
 #include "isothermalWall.hpp"
 #include "finiteVolume/compressibleFlowFields.hpp"
+#include "mathFunctions/functionFactory.hpp"
+#include "solver/dynamicRange.hpp"
 #include "utilities/mathUtilities.hpp"
 
 using fp = ablate::finiteVolume::CompressibleFlowFields;
 
-ablate::boundarySolver::lodi::IsothermalWall::IsothermalWall(std::shared_ptr<eos::EOS> eos, std::shared_ptr<finiteVolume::processes::PressureGradientScaling> pressureGradientScaling)
+ablate::boundarySolver::lodi::IsothermalWall::IsothermalWall(std::shared_ptr<eos::EOS> eos, std::shared_ptr<finiteVolume::processes::PressureGradientScaling> pressureGradientScaling,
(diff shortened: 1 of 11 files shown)


src/boundarySolver/lodi/isothermalWall.cpp
```diff
@@ -1,5 +1,7 @@
 #include "isothermalWall.hpp"
 #include "finiteVolume/compressibleFlowFields.hpp"
+#include "mathFunctions/functionFactory.hpp"
```
mmcgurn, line 3: Are these includes used?

src/boundarySolver/physics/sublimation.cpp
```diff
@@ -1,13 +1,17 @@
 #include "sublimation.hpp"
 #include <utility>
+#include "eos/radiationProperties/zimmer.hpp"
```
mmcgurn, line 3: why does this specific property header file need to be included?

src/boundarySolver/physics/sublimation.hpp
```diff
@@ -31,6 +33,8 @@ class Sublimation : public BoundaryProcess {
     // the pgs is needed for the pressure calculation
     const std::shared_ptr<finiteVolume::processes::PressureGradientScaling> pressureGradientScaling;
 
+    std::shared_ptr<ablate::radiation::Radiation> radiation;
```
mmcgurn, line 36: Would you mind improving the this header file with doxygen comments?

src/boundarySolver/physics/sublimation.cpp
```diff
@@ -261,6 +287,12 @@ void ablate::boundarySolver::physics::Sublimation::Setup(PetscInt numberSpeciesI
     computePressure = eos->GetThermodynamicTemperatureFunction(eos::ThermodynamicProperty::Pressure, {});
 }
 
+PetscErrorCode ablate::boundarySolver::physics::Sublimation::SublimationPreStep(TS ts, ablate::solver::Solver &solver) {
+    PetscFunctionBegin;
+    if (radiation) radiation->Solve(solver.GetSubDomain().GetSolutionVector(), solver.GetSubDomain().GetField("temperature"), solver.GetSubDomain().GetAuxVector());
```
mmcgurn, line 292: I don't think you need the radiation if test here.  This function should never be called if radiation is null

mmcgurn (changes requested):
@owenmylotte The PR looks good, I just have a few small comments.  Also, can you add the radiation heat flux to the SublimationOutputFunction so that it is recorded in the monitor?

src/boundarySolver/lodi/isothermalWall.hpp
```diff
@@ -2,6 +2,7 @@
 #define ABLATELIBRARY_ISOTHERMALWALL_HPP
 
 #include "lodiBoundary.hpp"
+#include "radiation/radiation.hpp"
```
mmcgurn, line 5: can you remove this?

src/boundarySolver/physics/CMakeLists.txt
```diff
@@ -7,4 +7,3 @@ target_sources(ablateLibrary
         sublimation.hpp
```
mmcgurn, line 7: and this extra space?

mmcgurn (changes requested):

mmcgurn (approved):

PR #348, opened 2022-09-20
Fixed TurbFlowStats

jpdening:


mmcgurn (changes requested):
Hi @jpdening.  Thanks for this fix, does it take care of the memory issue?  You will need to bump the version number.  I also thing that it is worth refactoring to work with the FieldMonitor abstract class.  Can you make that change for this PR?

src/monitors/turbFlowStats.cpp
```diff
@@ -26,253 +21,172 @@ PetscErrorCode ablate::monitors::TurbFlowStats::MonitorTurbFlowStats(TS ts, Pets
...
-        const auto& field = this->GetSolver()->GetSubDomain().GetField(fieldNames[f]);
-
-        fieldComps[f] = field.numberComponents;
-        numComp += field.numberComponents;
-        if ((f + 1) < fieldTrack.size()) {
-            fieldTrack[f + 1] = fieldTrack[f] + field.numberComponents;
+        const auto& field = solverIn->GetSubDomain().GetField(fieldNames[f]);
+        std::vector<std::string> innerCompNames(8 * field.numberComponents);
+        for (PetscInt c = 0; c < field.numberComponents; c++) {
+            for (std::size_t p = 0; p < suffix.size(); p++) {
```
mmcgurn, line 165: Would you mind removing the hard code 8 and 2.  If these are based upon the enums in the header file you can do something like https://stackoverflow.com/questions/2102582/how-can-i-count-the-items-in-an-enum

mmcgurn (changes requested):
The changes look great!  I just made one small request.  Thanks!

mmcgurn (approved):

PR #356, opened 2022-09-27
Made TurbFlowStats components uniquely named

jpdening:


src/monitors/turbFlowStats.cpp
```diff
@@ -163,7 +163,7 @@ void ablate::monitors::TurbFlowStats::Register(std::shared_ptr<ablate::solver::S
         std::vector<std::string> innerCompNames(SectionLabels::END * field.numberComponents);
         for (PetscInt c = 0; c < field.numberComponents; c++) {
             for (std::size_t p = 0; p < suffix.size(); p++) {
-                innerCompNames[SectionLabels::END * c + p] = field.name + suffix[p];
+                innerCompNames[SectionLabels::END * c + p] = field.name + "_" + field.components[c] + suffix[p];
```
mmcgurn, line 166: Thanks for making this change!  What do you think about add a wrap so it only adds this component if the field.numberComponents is > 1 or something similar
```c++
innerCompNames[SectionLabels::END * c + p] = field.name +(field.numberComponents > 1? ("_" + field.components[c]: "") + suffix[p]
```

mmcgurn (changes requested):
Thanks! Look good, I just had one thought.  You will also need to bump the version number.

mmcgurn (approved):
Thanks!  We might have to update the expected xdmf file if the variable names changed.

PR #360, opened 2022-10-04
fixed bug with mpi and boundary solver

mmcgurn:
fixed bug reported by @owenmylotte


mmcgurn:
> Are there some instances where an empty boundary array or face array are valid?

Over most of the domain both the boundaryArray and faceArray will be null but the cell range will also be zero.  I think that this is a result from the face in one region being next to a cell in another region.

PR #361, opened 2022-10-04
Update TurbFlowStats to work with a delay

jpdening:


src/monitors/turbFlowStats.cpp
```diff
@@ -120,8 +125,8 @@ PetscErrorCode ablate::monitors::TurbFlowStats::MonitorTurbFlowStats(TS ts, Pets
                         monitorPt[offset + SectionLabels::sumSqr] += fieldPt[p] * fieldPt[p];
                         monitorPt[offset + SectionLabels::favreAvg] =
                             monitorPt[offset + SectionLabels::densityDtMult] / (monitorPt[monitorFields[FieldPlacements::densityDtSum].offset] + Constant::tiny);
-                        monitorPt[offset + SectionLabels::rms] =
-                            PetscSqrtReal(monitorPt[offset + SectionLabels::sumSqr] / (step + Constant::tiny) - PetscPowReal(monitorPt[offset + SectionLabels::sum] / (step + Constant::tiny), 2));
+                        monitorPt[offset + SectionLabels::rms] = PetscSqrtReal(monitorPt[offset + SectionLabels::sumSqr] / (monitor->step + Constant::tiny) -
+                                                                               PetscPowReal(monitorPt[offset + SectionLabels::sum] / (monitor->step + Constant::tiny), 2));
```
mmcgurn, line 129: Why did this work before?  With an interval (even without the delay) the step did not match the number of times this was updated.

src/monitors/turbFlowStats.hpp
```diff
@@ -20,6 +20,7 @@ class TurbFlowStats : public FieldMonitor {
     const std::shared_ptr<ablate::eos::EOS> eos;
     const std::shared_ptr<io::interval::Interval> interval;
     ttf densityFunc;
+    PetscInt step;
```
mmcgurn, line 23: How is this step being saved/restored upon restart?

mmcgurn (approved):
thanks for making these changes!

PR #367, opened 2022-10-12
added sum and example radiation model

mmcgurn:
@owenmylotte and @klbudzin This should add the ability to have both soot and another absorptivity model


owenmylotte (approved):
Thank you!

PR #371, opened 2022-11-01
move volume fraction from aux to input field

maesemen:


src/finiteVolume/processes/twoPhaseEulerAdvection.cpp
```diff
@@ -203,29 +203,99 @@ ablate::finiteVolume::processes::TwoPhaseEulerAdvection::TwoPhaseEulerAdvection(
...
+    auto multiphasePreStage = std::bind(&ablate::finiteVolume::processes::TwoPhaseEulerAdvection::MultiphaseFlowPreStage, this, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3);
+    flow.RegisterPreStage(multiphasePreStage);
+
     // Create the decoder based upon the eoses
     decoder = CreateTwoPhaseDecoder(flow.GetSubDomain().GetDimensions(), eosGas, eosLiquid);
 
     // Currently, no option for species advection
-    flow.RegisterRHSFunction(CompressibleFlowComputeEulerFlux, this, "euler", {"densityVF", "euler"}, {});
-    flow.RegisterRHSFunction(CompressibleFlowComputeVFFlux, this, "densityVF", {"densityVF", "euler"}, {});
+    flow.RegisterRHSFunction(CompressibleFlowComputeEulerFlux, this, "euler", {"volumeFraction", "densityVF", "euler"}, {});
```
mmcgurn, line 214: Would you mind changing these to a const variable like CompressibleFlowFields::EULER_FIELD?

src/finiteVolume/processes/twoPhaseEulerAdvection.cpp
```diff
@@ -203,29 +203,99 @@ ablate::finiteVolume::processes::TwoPhaseEulerAdvection::TwoPhaseEulerAdvection(
...
+}
+PetscErrorCode ablate::finiteVolume::processes::TwoPhaseEulerAdvection::MultiphaseFlowPreStage(TS flowTs, ablate::solver::Solver &solver, PetscReal stagetime) {
+    PetscFunctionBegin;
+    // Get flow field data
+    const auto &fvSolver = dynamic_cast<ablate::finiteVolume::FiniteVolumeSolver &>(solver);
+    solver::Range cellRange;
+    fvSolver.GetCellRangeWithoutGhost(cellRange);
+    PetscInt dim;
+    PetscCall(DMGetDimension(fvSolver.GetSubDomain().GetDM(), &dim));
+    const auto &flowEulerId = fvSolver.GetSubDomain().GetField("euler").id;  // constant instead of string for euler
```
mmcgurn, line 238: Can you update this as well?

src/finiteVolume/processes/twoPhaseEulerAdvection.cpp
```diff
@@ -203,29 +203,99 @@ ablate::finiteVolume::processes::TwoPhaseEulerAdvection::TwoPhaseEulerAdvection(
...
+    auto &vfEulerId = fvSolver.GetSubDomain().GetField("volumeFraction").id;
+    DM dm = fvSolver.GetSubDomain().GetDM();
+    Vec globFlowVec;
+    PetscCall(TSGetSolution(flowTs, &globFlowVec));
+
+    const PetscScalar *flowArray;
+    PetscScalar *flowArrayV2;
+    PetscCall(VecGetArrayRead(globFlowVec, &flowArray));
```
mmcgurn, line 246: Why do you pull the flowArray twice?

src/finiteVolume/processes/twoPhaseEulerAdvection.cpp
```diff
@@ -203,29 +203,99 @@ ablate::finiteVolume::processes::TwoPhaseEulerAdvection::TwoPhaseEulerAdvection(
...
+        for (PetscInt d = 0; d < dim; d++) {
+            velocity[d] = eulerField[ablate::finiteVolume::CompressibleFlowFields::RHOU + d] / density;
+        }
+
+        PetscScalar *vfField = nullptr;
+        DMPlexPointLocalFieldRef(dm, cell, vfEulerId, flowArrayV2, &vfField) >> checkError;
+
+        // reproducing uOff
+        PetscInt uOff[3];
+        uOff[0] = 3 + dim;  // alpha
```
mmcgurn, line 264: I don't know if this is always true.  I think we need to get this from the fields.

tests/integrationTests/inputs/shocktube/shockTube1DSod_AirWater.yaml
```diff
@@ -43,12 +42,13 @@ timestepper:
     - &2
       fieldName: densityVF
       field: "x < 0.5 ? 11307.222318647702 : 0.0"
+    - &5
```
mmcgurn, line 45: Would you changing these anchors to be names instead of numbers?

tests/integrationTests/inputs/volumeOfFluids/twoGasAdvectingDiscontinuity.yaml
```diff
@@ -17,8 +17,8 @@ timestepper:
     upper: [ 10 ]
     boundary: [ NONE ]
     modifiers:
-      - !<!ablate::domain::modifiers::GhostBoundaryCells> ""
       - !<!ablate::domain::modifiers::DistributeWithGhostCells> ""
+      - !<!ablate::domain::modifiers::GhostBoundaryCells> ""
```
mmcgurn, line 21: Would you mind updating this yaml file to get rid of the extra !< around each tag and using names instead of numbers for the anchors?

mmcgurn (changes requested):
Thanks for starting this.  I have a few small requests and a question of uOff.  Are you still calling DecodeTwoPhaseEulerState when computing the fluxes? Does this recompute alpha still, or is that change part of another PR?

jikembo (approved):
Good change moving volume fraction to input field so that EOS can do its job.

src/finiteVolume/processes/twoPhaseEulerAdvection.cpp
```diff
@@ -203,29 +203,103 @@ ablate::finiteVolume::processes::TwoPhaseEulerAdvection::TwoPhaseEulerAdvection(
...
+    DM dm = fvSolver.GetSubDomain().GetDM();
+    Vec globFlowVec;
+    PetscCall(TSGetSolution(flowTs, &globFlowVec));
+
+    PetscScalar *flowArray;
+    PetscCall(VecGetArray(globFlowVec, &flowArray));
+
+    // reproducing uOff, based on fieldId
+    PetscInt rhoAlphaId = 0, vfId = 0;
+    if (rhoAlphaEulerId > flowEulerId) {
```
mmcgurn, line 251: These id orders are not always fixed and rhoAlphaEulerId may be before "Euler".  You should just pull the offset value directly from the field.

`PetscInt uOff[3] = {fvSolver.GetSubDomain().GetField("euler").offset, ..., ... }`

src/finiteVolume/processes/twoPhaseEulerAdvection.cpp
```diff
@@ -203,29 +203,103 @@ ablate::finiteVolume::processes::TwoPhaseEulerAdvection::TwoPhaseEulerAdvection(
...
+    PetscFunctionBegin;
+    // Get flow field data
+    const auto &fvSolver = dynamic_cast<ablate::finiteVolume::FiniteVolumeSolver &>(solver);
+    solver::Range cellRange;
+    fvSolver.GetCellRangeWithoutGhost(cellRange);
+    PetscInt dim;
+    PetscCall(DMGetDimension(fvSolver.GetSubDomain().GetDM(), &dim));
+    const auto &flowEulerId = fvSolver.GetSubDomain().GetField("euler").id;  // need this to get uOff
+    const auto &vfEulerId = fvSolver.GetSubDomain().GetField("volumeFraction").id;
+    const auto &rhoAlphaEulerId = fvSolver.GetSubDomain().GetField("densityVF").id;
```
mmcgurn, line 240: Can you use the variables such as CompressibleFlowFields::EULER_FIELD instead of strings here?

src/finiteVolume/processes/twoPhaseEulerAdvection.cpp
```diff
@@ -724,8 +766,10 @@ void ablate::finiteVolume::processes::TwoPhaseEulerAdvection::PerfectGasPerfectG
                                                                                                                     PetscReal *aG, PetscReal *aL, PetscReal *MG, PetscReal *ML, PetscReal *p,
                                                                                                                     PetscReal *T, PetscReal *alpha) {
     // (RHO, RHOE, RHOU, RHOV, RHOW)
-    const int EULER_FIELD = 1;
-    const int VF_FIELD = 0;
+    const int EULER_FIELD = 2;
+    const int VF_FIELD = 1;
+    // const int ALPHA_FIELD = 0; // new field
```
mmcgurn, line 771: If ALPHA_Field is not used can we remove it?

mmcgurn (changes requested):
Thanks for makes these changes.  I'm not sure if uOff is going to work but I added an example of something that I think will.

tests/unitTests/finiteVolume/processes/twoPhaseEulerAdvectionTests.cpp
```diff
@@ -62,7 +62,7 @@ TEST_P(TwoPhaseEulerAdvectionTestDecodeStateFixture, ShouldDecodeState) {
     PetscReal alpha;
 
     // compute offsets
-    PetscInt uOff[2] = {2 + params.dim /*densityYi*/, 0 /*euler*/};
+    PetscInt uOff[3] = {3 + params.dim /*alpha*/, 2 + params.dim /*rho1alpha1*/, 0 /*euler*/};
```
mmcgurn, line 65: It may be worth doing another test where you change the order of the fields

mmcgurn (changes requested):
Thanks for making these changes, I have just a few more comments.  How long does water gravity take to run?  We may want to use it as a regression test instead of integration so it can be larger.

mmcgurn (approved):

PR #374, opened 2022-11-02
update to the build presets/directions 

mmcgurn:
this update to the build presets/directions (on the wiki) should fix a build issue on linux.


owenmylotte (approved):
This works for me!

PR #376, opened 2022-11-07
tchem update

mmcgurn:
This PR switches to the Sandia repo for Tines and updates TChem to the latest CHREST version which allows for stream based echo when using Yaml input files.  This removes the need to round robin the initialization for tchem. As part of this new logs were added and the ability to get a stream from the log is now supported.  @Rozie100 this PR also cleans up some compiler warnings in the LES tests.  @jpdening would you mind building with this and running a case on quartz to see if there are any issues with large initialization?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,7 @@ if (NOT DEFINED CMAKE_CXX_COMPILER)
 endif ()
 
 # Set the project details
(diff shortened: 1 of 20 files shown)


Rozie100 (approved):
Sounds good to me.

jpdening (approved):
Initialization ran on a large 3D slab burner case

PR #381, opened 2022-11-14
update rocketMonitor output and integration test

salamlobad:
- Made rocket monitor output mDot in [x, y, z] directions.
- Made rocket monitor output cell pressure and mach number [min, max, mean] along faces
- Updated integration test for rocket monitor

### DIFF
--- a/src/monitors/rocketMonitor.cpp
+++ b/src/monitors/rocketMonitor.cpp
@@ -19,7 +19,7 @@ ablate::monitors::RocketMonitor::RocketMonitor(const std::string nameIn, std::sh
 
 void ablate::monitors::RocketMonitor::Register(std::shared_ptr<solver::Solver> solver) {
     Monitor::Register(solver);
-    auto comm = GetSolver()->GetSubDomain().GetComm();                                                                                       // The communicator in which the reduction takes place.
+    auto comm = GetSolver()->GetSubDomain().GetComm();                                                                                           // The communicator in which the reduction takes place.
     computePressure = eos->GetThermodynamicFunction(eos::ThermodynamicProperty::Pressure, GetSolver()->GetSubDomain().GetFields());          // get decode state function/context
     computeSpeedOfSound = eos->GetThermodynamicFunction(eos::ThermodynamicProperty::SpeedOfSound, GetSolver()->GetSubDomain().GetFields());  // get decode state function/context
     if (!log->Initialized()) {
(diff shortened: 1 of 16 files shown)


src/monitors/rocketMonitor.cpp
```diff
@@ -19,8 +19,9 @@ ablate::monitors::RocketMonitor::RocketMonitor(const std::string nameIn, std::sh
 
 void ablate::monitors::RocketMonitor::Register(std::shared_ptr<solver::Solver> solver) {
     Monitor::Register(solver);
-    auto comm = GetSolver()->GetSubDomain().GetComm();                                                                               // The communicator in which the reduction takes place.
-    computePressure = eos->GetThermodynamicFunction(eos::ThermodynamicProperty::Pressure, GetSolver()->GetSubDomain().GetFields());  // get decode state function/context
+    auto comm = GetSolver()->GetSubDomain().GetComm();                                                                                       // The communicator in which the reduction takes place.
+    computePressure = eos->GetThermodynamicFunction(eos::ThermodynamicProperty::Pressure, GetSolver()->GetSubDomain().GetFields());          // get decode state function/context
```
mmcgurn, line 23: If we are calling both of these in the same spot, we probably want to get the temperature first and then use that temperature to compute pressure and speed of sound.

src/monitors/rocketMonitor.cpp
```diff
@@ -81,6 +84,23 @@ PetscErrorCode ablate::monitors::RocketMonitor::OutputRocket(TS ts, PetscInt ste
         PetscReal thrustTotalGlob[3] = {0, 0, 0};
         PetscReal IspGlob[3] = {0, 0, 0};
 
+        // initialize field values to calculate [min max mean] for cell pressure and mach number
+        PetscReal pressureMin[1] = {100000000};
```
mmcgurn, line 88: I would use PETSC_MAX_REAL and PETSC_MIN_REAL just to be safe here.

src/monitors/rocketMonitor.cpp
```diff
@@ -126,30 +146,71 @@ PetscErrorCode ablate::monitors::RocketMonitor::OutputRocket(TS ts, PetscInt ste
...
+                            pressureMin[0] = cellPressure;
+                        }
+                        if (cellPressure > pressureMax[0]) {
+                            pressureMax[0] = cellPressure;
+                        }
+                        // update cell pressure sum for mean calculation
+                        pressureTotal[0] += cellPressure;
+
+                        // calculate cell mach number
+                        cellVelocity = PetscSqrtReal(PetscSqr(cellEuler[finiteVolume::CompressibleFlowFields::RHOU]) + PetscSqr(cellEuler[finiteVolume::CompressibleFlowFields::RHOU + 1]) +
```
mmcgurn, line 160: Will this work in 1 and 2D?

src/monitors/rocketMonitor.cpp
```diff
@@ -81,6 +84,23 @@ PetscErrorCode ablate::monitors::RocketMonitor::OutputRocket(TS ts, PetscInt ste
...
         PetscReal IspGlob[3] = {0, 0, 0};
 
+        // initialize field values to calculate [min max mean] for cell pressure and mach number
+        PetscReal pressureMin[1] = {100000000};
+        PetscReal pressureMax[1] = {0};
+        PetscReal pressureTotal[1] = {0};
+        PetscReal pressureMinGlob[1];
+        PetscReal pressureMaxGlob[1];
+        PetscReal pressureTotalGlob[1];
+        PetscReal machMin[1] = {100000000};
```
mmcgurn, line 94: This do not need to be arrays to send them over mpi.  You can send scalars over mpi by giving the memory address of the scalar as &varName

src/monitors/rocketMonitor.cpp
```diff
@@ -126,30 +146,71 @@ PetscErrorCode ablate::monitors::RocketMonitor::OutputRocket(TS ts, PetscInt ste
...
+                        // update mach number sum for mean calculation
+                        machTotal[0] += machNumber;
+                        numCells[0]++;  // Track number of cells on face in region of interest
                     }
                 }
             }
         }
 
         // Take across all ranks
+        MPI_Reduce(pressureMin, pressureMinGlob, 1, MPI_DOUBLE, MPI_MIN, 0, comm);
```
mmcgurn, line 181: These are a lot of calls to MPI_Reduce.  We might want to instead create a few arrays and store pressure, mach ... into a single array.

mmcgurn (changes requested):
I have some small comments we can talk about tomorrow.  Thanks for putting this together.

PR #386, opened 2022-11-15
equation interval

### LINKED ISSUE #385: Interval with Equation
An interval class that takes an equation as an input.

mmcgurn:
This should close #385


owenmylotte (approved):
This will be very helpful. Thank you!

PR #391, opened 2022-11-22
Stage Losses Update

owenmylotte:
Moves the radiation losses to the pre-stage function to provide time stepping stability. The radiation gains are evaluated at the interval. This allows the volumetric radiation to run stable with mass fraction dependent absorption.

### DIFF
--- a/src/boundarySolver/physics/sublimation.cpp
+++ b/src/boundarySolver/physics/sublimation.cpp
@@ -91,7 +91,7 @@ void ablate::boundarySolver::physics::Sublimation::Initialize(ablate::boundarySo
         solver::DynamicRange faceRange;
         for (const auto &i : bSolver.GetBoundaryGeometry()) {
             PetscInt ghost = -1;
-            DMLabelGetValue(ghostLabel, i.geometry.faceId, &ghost) >> checkError;
+            if (ghostLabel) DMLabelGetValue(ghostLabel, i.geometry.faceId, &ghost) >> checkError;
             if (!(ghost >= 0)) faceRange.Add(i.geometry.faceId);  //!< Add each ID to the range that the radiation solver will use
         }
         radiation->Setup(faceRange.GetRange(), bSolver.GetSubDomain());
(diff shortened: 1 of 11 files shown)


owenmylotte (commented):
The only substantial change is that the calls to the solve and the radiation gain evaluation are separated.
All of the current tests pass, but the optimized build segfault is still present in the slab burner. I'll try running some additional simple cases to isolate it. (I've gotten the source down to the sublimation solver absorption function call at this point, maybe an initialization issue). It may be worth adding additional integration tests for different radiation and absorption classes to test that all of the solver interactions are working?
Once the optimized build issue is figured out then I'm confident we have a running full radiation slab burner.

src/boundarySolver/physics/sublimation.cpp
```diff
@@ -295,15 +301,16 @@ void ablate::boundarySolver::physics::Sublimation::Setup(PetscInt numberSpeciesI
...
-PetscErrorCode ablate::boundarySolver::physics::Sublimation::SublimationPreStep(BoundarySolver &solver, TS ts, PetscReal time, bool initialStage, Vec locX, void *ctx) {
+PetscErrorCode ablate::boundarySolver::physics::Sublimation::SublimationPreRHS(BoundarySolver &solver, TS ts, PetscReal time, bool initialStage, Vec locX, void *ctx) {
     PetscFunctionBegin;
     auto sublimation = (Sublimation *)ctx;
     PetscInt step;
     TSGetStepNumber(ts, &step) >> checkError;
     TSGetTime(ts, &time) >> checkError;
     if (initialStage && sublimation->radiationInterval->Check(PetscObjectComm((PetscObject)ts), step, time)) {
-        sublimation->radiation->Solve(solver.GetSubDomain().GetSolutionVector(), solver.GetSubDomain().GetField("temperature"), solver.GetSubDomain().GetAuxVector());
+        sublimation->radiation->EvaluateGains(solver.GetSubDomain().GetSolutionVector(), solver.GetSubDomain().GetField("temperature"), solver.GetSubDomain().GetAuxVector());
```
mmcgurn, line 311: This is a good name change.

src/eos/radiationProperties/zimmer.cpp
```diff
@@ -198,6 +220,8 @@ ablate::eos::ThermodynamicTemperatureFunction ablate::eos::radiationProperties::
                                     .densityYiCO2Offset = CO2offset,
                                     .densityYiCOOffset = COoffset,
                                     .densityYiCH4Offset = CH4offset,
+                                    .eulerOffset = eulerField->offset,
```
mmcgurn, line 223: Where do you use this Euler offset?

mmcgurn (approved):
I have a question about the added eulerOffset to the zimmer model.

PR #397, opened 2022-12-06
fixed bug in mixture fraction monitor when used with a non-subdomain

mmcgurn:


owenmylotte (approved):
Checking that the pointer exists make sense. Also, the check for the pointer when taking a value from the array is reminiscent of the iteration through CellRange.

PR #398, opened 2022-12-06
Memory Fix, Volumetric Limiter

owenmylotte:
Reduces the memory consumption of the radiation solver and protects the volumetric radiation from eos integration failures. Additional changes to enforce positivity in the absorption model.

### DIFF
--- a/src/eos/radiationProperties/zimmer.cpp
+++ b/src/eos/radiationProperties/zimmer.cpp
@@ -64,7 +64,11 @@ PetscErrorCode ablate::eos::radiationProperties::Zimmer::ZimmerFunction(const Pe
         pCO = (density * UGC * YinCO * temperature) / (MWCO * 101325.);
 
         /** The resulting absorptivity is an average of species absorptivity weighted by partial pressure. */
-        *kappa = pCO2 * kappaCO2 + pH2O * kappaH2O + pCH4 * kappaCH4 + pCO * kappaCO;
+        *kappa = 0;
+        if (pCO2 > 1E-3 && kappaCO2 > 0) *kappa += pCO2 * kappaCO2;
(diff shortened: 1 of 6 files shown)


src/eos/radiationProperties/zimmer.cpp
```diff
@@ -64,7 +64,11 @@ PetscErrorCode ablate::eos::radiationProperties::Zimmer::ZimmerFunction(const Pe
         pCO = (density * UGC * YinCO * temperature) / (MWCO * 101325.);
 
         /** The resulting absorptivity is an average of species absorptivity weighted by partial pressure. */
-        *kappa = pCO2 * kappaCO2 + pH2O * kappaH2O + pCH4 * kappaCH4 + pCO * kappaCO;
+        *kappa = 0;
```
owenmylotte, line 67: This enforces positive values out of the absorption model.

src/radiation/radiation.cpp
```diff
@@ -738,7 +724,6 @@ void ablate::radiation::Radiation::EvaluateGains(Vec solVec, ablate::domain::Fie
     VecRestoreArrayRead(solVec, &solArray);
     VecRestoreArrayRead(auxVec, &auxArray);
     VecRestoreArrayRead(faceGeomVec, &faceGeomArray) >> checkError;
-    EndEvent();
```
mmcgurn, line 741: It looks like after the rebase we may be missing some Start/End events

mmcgurn (changes requested):

src/radiation/radiation.cpp
mmcgurn: I added back the events and pushed to the PR branch.

owenmylotte:
Thank you for fixing this!

mmcgurn:
This will be replaced with an updated PR using PetscSF for rad.

PR #401, opened 2022-12-09
added high level logging to finite volume solver/domain init

### LINKED ISSUE #399: Profiling Initialization
@mmcgurn ,
  Please let me know when you have added the code allow for profiling initialization.

Thanks
Al

mmcgurn:
@arcowie I think this should close #399 .  @owenmylotte would you mind taking a look to see if this change is ok in the radiation solver?

### DIFF
--- a/src/finiteVolume/finiteVolumeSolver.cpp
+++ b/src/finiteVolume/finiteVolumeSolver.cpp
@@ -9,7 +9,7 @@
 ablate::finiteVolume::FiniteVolumeSolver::FiniteVolumeSolver(std::string solverId, std::shared_ptr<domain::Region> region, std::shared_ptr<parameters::Parameters> options,
                                                              std::vector<std::shared_ptr<processes::Process>> processes,
                                                              std::vector<std::shared_ptr<boundaryConditions::BoundaryCondition>> boundaryConditions, bool computePhysicsTimeStep)
-    : CellSolver(std::move(solverId), std::move(region), std::move(options)),
+    : CellSolver(solverId, std::move(region), std::move(options)),
       computePhysicsTimeStep(computePhysicsTimeStep),
       processes(std::move(processes)),
       boundaryConditions(std::move(boundaryConditions)),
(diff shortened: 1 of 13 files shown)


src/finiteVolume/finiteVolumeSolver.cpp
```diff
@@ -9,7 +9,7 @@
 ablate::finiteVolume::FiniteVolumeSolver::FiniteVolumeSolver(std::string solverId, std::shared_ptr<domain::Region> region, std::shared_ptr<parameters::Parameters> options,
                                                              std::vector<std::shared_ptr<processes::Process>> processes,
                                                              std::vector<std::shared_ptr<boundaryConditions::BoundaryCondition>> boundaryConditions, bool computePhysicsTimeStep)
-    : CellSolver(std::move(solverId), std::move(region), std::move(options)),
+    : CellSolver(solverId, std::move(region), std::move(options)),
```
owenmylotte, line 12: What is the effect of removing this?

src/radiation/radiation.hpp
```diff
@@ -13,8 +13,8 @@
 
 namespace ablate::radiation {
 
-class Radiation : public utilities::Loggable<Radiation> {  //!< Cell solver provides cell based functionality, right hand side function compatibility with
-                                                           //!< finite element/ volume, loggable allows for the timing and tracking of events
+class Radiation : protected utilities::Loggable<Radiation> {  //!< Cell solver provides cell based functionality, right hand side function compatibility with
```
owenmylotte, line 16: This makes sense.

src/solver/timeStepper.cpp
```diff
@@ -57,6 +57,7 @@ ablate::solver::TimeStepper::TimeStepper(std::string nameIn, std::shared_ptr<abl
 ablate::solver::TimeStepper::~TimeStepper() { TSDestroy(&ts) >> checkError; }
 
 void ablate::solver::TimeStepper::Initialize() {
+    StartEvent((this->name + "::Initialize").c_str());
```
owenmylotte, line 60: Will this duplicate the individual Initialize logs in the child classes or is this only acting on the domain setups?

owenmylotte (approved):
Thank you!

src/solver/timeStepper.cpp
mmcgurn: This will only act on the highest level timestpper init.  If any other solver records an init it will be under this one.

PR #402, opened 2022-12-12
BoxMesh

Rozie100:
3D BoxMesh values are updated.

### DIFF
--- a/src/domain/boxMesh.cpp
+++ b/src/domain/boxMesh.cpp
@@ -61,10 +61,10 @@ REGISTER(ablate::domain::Domain, ablate::domain::BoxMesh,
          "\n### 3D:\n\n"
          "| Direction | Description | Value |\n"
          "| --- | --- |-----|\n"
-         " | y- | bottom | 1 |\n"
-         " | y+ | top    | 2|\n"
-         " | z+ | front  | 3|\n"
-         " | z- | back   | 4|\n"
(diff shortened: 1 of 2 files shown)


src/domain/boxMesh.cpp
```diff
@@ -61,14 +61,14 @@ REGISTER(ablate::domain::Domain, ablate::domain::BoxMesh,
          "\n### 3D:\n\n"
          "| Direction | Description | Value |\n"
          "| --- | --- |-----|\n"
-         " | y- | bottom | 1 |\n"
-         " | y+ | top    | 2|\n"
-         " | z+ | front  | 3|\n"
-         " | z- | back   | 4|\n"
+         " | y- | bottom | 3 |\n"
```
mmcgurn, line 64: These were based upon https://gitlab.com/petsc/petsc/-/blob/main/src/dm/impls/plex/plexcreate.c#L867 so our definition of front/back/top might be different.  Would you mind looking at the code to check?  Also you have front/back as both 1

mmcgurn (changes requested):

src/domain/boxMesh.cpp
Rozie100: I see now. So in the code "front and back " are y+ and y-. So, I should change the definition of the directions equivalently.

mmcgurn (approved):

PR #404, opened 2022-12-14
Log law boundary

Rozie100:
A new log-law BC is added.
The turbulentFlow folder is generated for integration tests.


CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ if (CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
 endif ()
 
 # Create the library for ablate
-add_library(ablateLibrary SHARED)
+add_library(ablateLibrary SHARED src/boundarySolver/physics/logLawBoundary.cpp src/boundarySolver/physics/logLawBoundary.hpp)
```
mmcgurn, line 41: Can you add these files to the cmake list file in the physics folder instead?

src/boundarySolver/physics/logLawBoundary.cpp
```diff
@@ -0,0 +1,61 @@
...
+    PetscArrayzero(logLawVel, dim);
+    PetscReal stencilVel[dim];
+    PetscReal stencilNormalVelocity[dim];
+    PetscReal logLaw;
+    // get the stencil velocity in normal coord.
+    PetscReal stencilDensity = stencilValues[uOff[EULER_FIELD] + finiteVolume::CompressibleFlowFields::RHO];
+    for (PetscInt d = 0; d < dim; d++) {
+        stencilVel[d] = stencilValues[uOff[EULER_FIELD] + finiteVolume::CompressibleFlowFields::RHOU + d] / stencilDensity;
+        PetscReal transformationMatrix[3][3];
+        utilities::MathUtilities::ComputeTransformationMatrix(dim, fg->normal, transformationMatrix);
```
mmcgurn, line 26: Can we do this transformation outside of the dimension loop?

src/boundarySolver/physics/logLawBoundary.cpp
```diff
@@ -0,0 +1,61 @@
...
+        ablate::utilities::MathUtilities::Multiply(dim, transformationMatrix, stencilVel, stencilNormalVelocity);
+        // calculate the new boundary velocity in normal coord.
+        if (d == 1) {
```
mmcgurn, line 29: If we this outside of the dimension loop, we could get rid of the if statements.

src/boundarySolver/physics/logLawBoundary.cpp
```diff
@@ -0,0 +1,61 @@
...
+        if (d == 2) {
+            logLaw = stencilNormalVelocity[d];
+        }
+        logLawVel[d] += logLaw;
+    }
+    // transform to Cartesian coord.
+    PetscReal boundaryVel[dim];
+    PetscReal transformationMatrix[dim][3];
+    PetscReal velocityCartSystem[dim];
+    utilities::MathUtilities::ComputeTransformationMatrix(dim, fg->normal, transformationMatrix);
```
mmcgurn, line 47: Can we reduce the same transformation matrix as before?

tests/integrationTests/inputs/compressibleFlow/turbulentFlow/turbulentChannelFlow.yaml
```diff
@@ -0,0 +1,208 @@
+#0 This case is inspired by the setup for Direct Numerical Simulation of turbulent channel flow up to Ret=590 (Moser et al. Physics of Fluids, 11(4):943–945, 1999).
+# The case is a statistically-developing internal flow through parallel smooth walls.
+# The domain comprises a 3D rectangular channel.
+# Longitudinal direction (Mean-flow direction)
+# Vertical direction (Wall-normal direction)
+# Spanwise direction (Statistically homogeneous direction).
+# Large eddy simulation (LES) is used to model sub-grid-scales and wall models are deployed near the wall regions.
+
+# Turbulent inflow is imposed at the intel. Turbulent velocity profile is approximated from 1/7 power law i.e., u = Umax (y/Re)^1/7 with synthetic turbulence generator.
```
mmcgurn, line 9: I think that if you wrap the equations with &&, they will show up as equations in the docs, you can preview this using either https://ablate.dev/content/development/Documentation.html#local-docs-preview or https://ablate.dev/content/development/Documentation.html#pull-request-preview

tests/unitTests/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-add_executable(unitTests "")
+add_executable(unitTests "" boundarySolver/physics/logLawBoundaryTests.cpp)
```
mmcgurn, line 1: can you remove this if it already added to the other cmakeLists?

mmcgurn (changes requested):
Thanks for adding this!  I have some comments on the dimension loop where I think we can save some computational cost.

tests/unitTests/CMakeLists.txt
Rozie100: Not intentional. Must have done something wrong.

src/boundarySolver/physics/logLawBoundary.cpp
```diff
@@ -0,0 +1,58 @@
...
+                   fg->normal[1] * 0.5 * stencilNormalVelocity[1];
+
+    logLawVel[0] = 0e+0;
+    logLawVel[2] = stencilNormalVelocity[2];
+
+    // transform back to Cartesian coord.
+    PetscReal boundaryVel[dim];
+    PetscReal velocityCartSystem[dim];
+
+    utilities::MathUtilities::ComputeTransformationMatrix(dim, fg->normal, transformationMatrix);
```
mmcgurn, line 43: Why do we need to recompute this?

mmcgurn (changes requested):

src/boundarySolver/physics/logLawBoundary.cpp
Rozie100: Not really.

Rozie100: Thanks for the comments and insights towards making it more efficient.

src/boundarySolver/physics/logLawBoundary.cpp
```diff
@@ -0,0 +1,57 @@
...
+    for (PetscInt d = 0; d < dim; d++) {
+        stencilVel[d] = stencilValues[uOff[EULER_FIELD] + finiteVolume::CompressibleFlowFields::RHOU + d] / stencilDensity;
+    }
+
+    PetscReal transformationMatrix[dim][3];
+    utilities::MathUtilities::ComputeTransformationMatrix(dim, fg->normal, transformationMatrix);
+    ablate::utilities::MathUtilities::Multiply(dim, transformationMatrix, stencilVel, stencilNormalVelocity);
+
+    // calculate the new boundary velocity in normal coord.
+    logLawVel[1] = -fg->normal[1] * -0.5 * (1 / kappa) * 2 / ((boundaryCell->volume * boundaryCell->volume / (fg->areas[1] * fg->areas[1])) * abs(fg->normal[1])) /
```
mmcgurn, line 32: Why does the tangent velocities only get updated in a single direction?

Rozie100: I should be changing both directions based on the Dominant velocity direction, also like you said the 2D is different.
My tests passed in the first round when I had a loop because it was picking the velocities consistently, although not in the correct direction necessarily.

src/boundarySolver/physics/logLawBoundary.cpp
```diff
@@ -0,0 +1,71 @@
...
+    ablate::utilities::MathUtilities::Multiply(dim, transformationMatrix, stencilVel, stencilNormalVelocity);
+
+    // calculate the new boundary velocity in normal coord for 2D.
+    logLawVel[0] = 0e+0;
+
+    logLawVel[1] = -fg->normal[1] * 0.5 * (1 / kappa) * 2 / ((boundaryCell->volume * boundaryCell->volume / (fg->areas[1] * fg->areas[1]))) /
+                       (4 / ((boundaryCell->volume * boundaryCell->volume / (fg->areas[1] * fg->areas[1]))) - 4 / (fg->areas[1])) +
+                   0.5 * stencilNormalVelocity[1];
+    // for 3D
+    if ((dim = 3)) {
```
mmcgurn, line 39: should this be dim ==3?

src/boundarySolver/physics/logLawBoundary.cpp
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    PetscReal transformationMatrix[dim][3];
+    utilities::MathUtilities::ComputeTransformationMatrix(dim, fg->normal, transformationMatrix);
+    ablate::utilities::MathUtilities::Multiply(dim, transformationMatrix, stencilVel, stencilNormalVelocity);
+
+    // calculate the new boundary velocity in normal coord for 2D.
+    logLawVel[0] = 0e+0;
+
+    logLawVel[1] = -fg->normal[1] * 0.5 * (1 / kappa) * 2 / ((boundaryCell->volume * boundaryCell->volume / (fg->areas[1] * fg->areas[1]))) /
```
mmcgurn, line 35: I think the areas are in the cartesian coordinate system, do we need to map that?

mmcgurn (changes requested):
Can you explain this log a little in the comments?  If?
vel_t2 > vel_t1
  logLawVel[1] = modified stencilNormalVelocity[1]
  logLawVel[2] = modified stencilNormalVelocity[2]
vel_t1 > vel_t2
  logLawVel[1] = modified stencilNormalVelocity[1]
  logLawVel[2] = stencilNormalVelocity[2]

I think that vel_t2 and vel_t1 are arbitrary, shouldn't the result be the same if the values are flipped?

src/boundarySolver/physics/logLawBoundary.cpp
Rozie100: My tests only had velocities in x-dir, that's why they did not catch the missing component. I added more tests for bi-directional velocities.

Now that I am calculating the new magnitude and mapping it, for a single direction of velocities I get the same results as before, however it makes a difference in multi-directional cases

src/boundarySolver/physics/logLawBoundary.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+    PetscReal transformationMatrix[dim][3];
+    utilities::MathUtilities::ComputeTransformationMatrix(dim, fg->normal, transformationMatrix);
+    ablate::utilities::MathUtilities::Multiply(dim, transformationMatrix, stencilVel, stencilNormalVelocity);
+    PetscReal area = utilities::MathUtilities::MagVector(dim, fg->areas);
+
+    // calculate the new boundary velocity in normal coord for 2D.
+    logLawVel[0] = 0e+0;
+
+    logLawVel[1] =
```
mmcgurn, line 39: Thanks for making these changes!  We could save a little work by wrapping this in a if dim == 2 else if dim == 3 statement.  Would you mind making this change?

mmcgurn (changes requested):

src/boundarySolver/physics/logLawBoundary.cpp
Rozie100: Makes sense!

mmcgurn (approved):

PR #405, opened 2022-12-24
Turb flow monitor

Rozie100:
The current monitor does not march over all components of the velocity field.

### DIFF
--- a/src/monitors/turbFlowStats.cpp
+++ b/src/monitors/turbFlowStats.cpp
@@ -112,12 +112,13 @@ PetscErrorCode ablate::monitors::TurbFlowStats::MonitorTurbFlowStats(TS ts, Pets
                     monitorPt[monitorFields[FieldPlacements::densitySum].offset] += densLoc;
                     monitorPt[monitorFields[FieldPlacements::densityDtSum].offset] += densLoc * dt;
 
-                    // Set the offset
-                    PetscInt offset = monitorFields[FieldPlacements::fieldsStart + f].offset;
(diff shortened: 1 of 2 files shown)


src/monitors/turbFlowStats.cpp
```diff
@@ -112,12 +112,13 @@ PetscErrorCode ablate::monitors::TurbFlowStats::MonitorTurbFlowStats(TS ts, Pets
                     monitorPt[monitorFields[FieldPlacements::densitySum].offset] += densLoc;
                     monitorPt[monitorFields[FieldPlacements::densityDtSum].offset] += densLoc * dt;
 
-                    // Set the offset
-                    PetscInt offset = monitorFields[FieldPlacements::fieldsStart + f].offset;
+                    const auto& field = monitor->GetSolver()->GetSubDomain().GetField(monitor->fieldNames[f]);
 
                     // March over each field component
-                    for (int p = 0; p < monitorFields[f].numberComponents; p++) {
-                        offset += SectionLabels::END * p;
```
mmcgurn, line 120: Was this only broken when dim > 2? This change makes sense to me but it took a while to figure out the math.  Can you add a comment that explains the offset and how the components are arranged?

mmcgurn (approved):
Thanks for catching this!  I added a comment about adding some comments.

src/monitors/turbFlowStats.cpp
Rozie100: Both 3D and 2D actually. The integration test, which is 2D gives garbage for the second velocity.

Rozie100: Tried to explain it. let me know if its is confusing

PR #408, opened 2022-12-31
adds support for new radiation solver directly using petscsf

mmcgurn:
This appears to help resolve swarmDm bottle neck in radiation solver:

Before:
image

After:
image


src/radiation/CMakeLists.txt
```diff
@@ -2,12 +2,10 @@ target_sources(ablateLibrary
         PRIVATE
         radiation.cpp
         volumeRadiation.cpp
-        raySharingRadiation.cpp
```
owenmylotte, line 5: Temporary elimination of ray sharing.

src/radiation/radiation.hpp
```diff
@@ -28,29 +30,29 @@ class Radiation : protected utilities::Loggable<Radiation> {  //!< Cell solver p
 
     virtual ~Radiation();
 
+    /** Identifiers are carrying by both the search and solve particles in order to associate them with their origins and ray segments
```
owenmylotte, line 33: Identifying information data structure change.

src/radiation/surfaceRadiation.cpp
```diff
@@ -21,83 +14,65 @@ void ablate::radiation::SurfaceRadiation::Initialize(const solver::Range& cellRa
     struct Identifier* identifier;      //!< Pointer to the ray identifier information
 
     /** Get the fields associated with the particle swarm so that they can be modified */
-    DMSwarmGetField(radsearch, DMSwarmPICField_coor, nullptr, nullptr, (void**)&coord) >> checkError;
-    DMSwarmGetField(radsearch, DMSwarmPICField_cellid, nullptr, nullptr, (void**)&index) >> checkError;
-    DMSwarmGetField(radsearch, "identifier", nullptr, nullptr, (void**)&identifier) >> checkError;
-    DMSwarmGetField(radsearch, "virtual coord", nullptr, nullptr, (void**)&virtualcoord) >> checkError;
+    DMSwarmGetField(radSearch, DMSwarmPICField_coor, nullptr, nullptr, (void**)&coord) >> checkError;
+    DMSwarmGetField(radSearch, DMSwarmPICField_cellid, nullptr, nullptr, (void**)&index) >> checkError;
+    DMSwarmGetField(radSearch, IdentifierField, nullptr, nullptr, (void**)&identifier) >> checkError;
```
owenmylotte, line 19: MPI struct replacement.

src/radiation/surfaceRadiation.cpp
```diff
@@ -21,83 +14,65 @@ void ablate::radiation::SurfaceRadiation::Initialize(const solver::Range& cellRa
...
     ablate::radiation::Radiation::Initialize(cellRange, subDomain);
+
+    // set up the reverse lookup for faces
+    indexLookup = solver::ReverseRange(cellRange);
 }
 
-PetscReal ablate::radiation::SurfaceRadiation::SurfaceComponent(DM faceDM, const PetscScalar* faceGeomArray, PetscInt iCell, PetscInt nphi, PetscInt ntheta) {
-    PetscFVFaceGeom* faceGeom;
-    DMPlexPointLocalRead(faceDM, iCell, faceGeomArray, &faceGeom) >> checkError;
+PetscReal ablate::radiation::SurfaceRadiation::SurfaceComponent(const PetscReal normal[], PetscInt iCell, PetscInt nphi, PetscInt ntheta) {
```
owenmylotte, line 60: Give normal instead of all face geometry.

src/radiation/volumeRadiation.cpp
```diff
@@ -54,15 +59,43 @@ PetscErrorCode ablate::radiation::VolumeRadiation::ComputeRHSFunction(PetscReal
...
+
+    /** Declare the basic information*/
+    PetscReal* sol = nullptr;          //!< The solution value at any given location
+    PetscReal* temperature = nullptr;  //!< The temperature at any given location
+    double kappa = 1;                  //!< Absorptivity coefficient, property of each cell
+
+    auto absorptivityFunctionContext = absorptivityFunction.context.get();  //!< Get access to the absorption function
+
+    //!< This will iterate only though local cells
+    auto& range = radiationCellRange.GetRange();
```
owenmylotte, line 83: Move solve functionality outside of the base class

src/solver/reverseRange.hpp
```diff
@@ -0,0 +1,55 @@
...
+    PetscInt pointStart = 0;
+    std::vector<PetscInt> indices;
+
+   public:
+    explicit ReverseRange(const Range& range) {
+        rangeStart = range.start;
+        if (range.points && (range.start != range.end)) {
+            // find the min/max point
+            PetscInt maxPoint = rangeStart;
+            for (PetscInt i = range.start; i < range.end; ++i) {
```
owenmylotte, line 23: Range index lookup

owenmylotte (approved):
Exciting!

src/radiation/radiation.cpp
```diff
@@ -364,81 +376,90 @@ void ablate::radiation::Radiation::UpdateCoordinates(PetscInt ipart, Virtualcoor
 }
 
 PetscReal ablate::radiation::Radiation::FaceIntersect(PetscInt ip, Virtualcoord* virtualcoord, PetscFVFaceGeom* faceGeom) const {
-    PetscReal ldotn = (virtualcoord[ip].xdir * faceGeom->normal[0]) + (virtualcoord[ip].ydir * faceGeom->normal[1]) + (virtualcoord[ip].zdir * faceGeom->normal[2]);
+    //!<(planeNormal.dot(planePoint) - planeNormal.dot(linePoint)) / planeNormal.dot(lineDirection.normalize())
+    PetscReal ldotn = 0.0;
```
mmcgurn, line 380: @owenmylotte I think we were using uninitialized data when 1 or 2 Dim.  Does this change make sense?

src/radiation/radiation.cpp
```diff
@@ -130,17 +118,23 @@ void ablate::radiation::Radiation::Setup(const solver::Range& cellRange, ablate:
                 virtualcoord[ipart].y = centroid[1] + (virtualcoord[ipart].ydir * 0.1 * minCellRadius);
                 virtualcoord[ipart].z = centroid[2] + (virtualcoord[ipart].zdir * 0.1 * minCellRadius);
 
+                // Init hhere to default value
```
mmcgurn, line 121: @owenmylotte I don't think we had a default value here, is this ok?

src/radiation/radiation.cpp
owenmylotte: I'm not sure if the switch makes sense to me because 2D needs to add x and y components while 3D needs to add x, y, and z. Will this use all of the components in 3D and I'm misunderstanding?

owenmylotte: Ah, I think I was missing the fall through. I haven't seen that before.

src/radiation/radiation.cpp
owenmylotte: That looks good to me. The path length should always be set before it's entered in the vector so having it uninitialized shouldn't cause problems. Zero is good. Was this causing a bug?

src/radiation/radiation.cpp
mmcgurn: You don't really need the fall through, but c/c++ does fall through by default (probably a bad design).  So you can annotate your code with fall through because most modern compilers now warn you about fall through if you don't annotate it.

src/radiation/radiation.cpp
mmcgurn: I'm not sure, either this or FaceIntersect was reporting using uninitialized data.

src/radiation/radiation.cpp
owenmylotte: Yes that looks good to me.

PR #409, opened 2023-01-04
Two Phase Equation of State

maesemen:
twoPhase equation of state and unit tests.  Still unsure of how getFieldFunctions should behave for this implementation.  Need wall boundary condition to do dam break integration test.


src/eos/twoPhase.cpp
```diff
@@ -0,0 +1,1075 @@
...
+    // simple decode to compute pressure
+    PetscReal derived[6];
+    PetscReal in[10];
+    in[0] = conserved[functionContext->volumeFractionOffset];
+    in[1] = conserved[functionContext->densityVFOffset];
+    in[2] = density;
+    in[3] = internalEnergy;
+    in[4] = functionContext->parameters.gamma1;
+    in[5] = functionContext->parameters.gamma2;
+    if (functionContext->parameters.p01 == 0 && functionContext->parameters.p02 == 0){
```
mmcgurn, line 641: I would recommend not doing this check everytime.  Instead have a SensibleEnthalpyFunction for each possible interaction that handles it directly.  Then return the specific SensibleEnthalpyFunction for the scenario.  If it is easier to think about you could create three equation of states and have this one just create the one needed.

tests/unitTests/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-add_executable(unitTests "")
+add_executable(unitTests "" eos/twoPhaseTests.cpp)
```
mmcgurn, line 1: Can you remove this from here?

mmcgurn (changes requested):
Thanks for making this PR.  I think we may want to simplify the equation of state by splitting up each of the functions into their specific version.  I add a little more about this comment around this code.  Please let me know if you have any questions.

src/eos/twoPhase.cpp
maesemen: I will need your help on how to change this. Still haven't come up with a way to do this efficiently, since this effects the P, T, h, Cv, Cp, and sound speed functions.

mmcgurn: I would have separate functions for each, something like:

- SensibleEnthalpyFunctionSimpleGasSimpleGas
- SensibleEnthalpyFunctionSimpleGasStiff
- SensibleEnthalpyFunctionStiffStiff

You could do this for each P, T, h, Cv, Cp.  The question is how much extra work is done in the decode state, would it be more efficient to move the parts of each needed decode into the separate P, T, ... functions?  If all of the decode needs to be done everytime and we should keep the function I would replace the array with indexes with a struct to make it easier to read.

mmcgurn (approved):

PR #412, opened 2023-01-06
TKE is bounded.

Rozie100:
TKE is bounded.

### DIFF
--- a/src/finiteVolume/processes/les.cpp
+++ b/src/finiteVolume/processes/les.cpp
@@ -69,7 +69,7 @@ PetscErrorCode ablate::finiteVolume::processes::LES::LesMomentumFlux(PetscInt di
 
     turbulence = aux[aOff[TKE_FIELD]];
     // Bound TKE field if it is gone unbounded.
-    turbulence = fmax(0, turbulence);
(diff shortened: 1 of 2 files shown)


src/finiteVolume/processes/les.cpp
```diff
@@ -68,6 +68,8 @@ PetscErrorCode ablate::finiteVolume::processes::LES::LesMomentumFlux(PetscInt di
     PetscReal turbulence;
 
     turbulence = aux[aOff[TKE_FIELD]];
+    // Bound TKE field if it is gone unbounded.
+    turbulence = fmax(0, turbulence);
```
mmcgurn, line 72: For petsc values you will want to user PetscMax instead of max.

mmcgurn (changes requested):
Hi!  You may want to use PetscMax instead because this is a petscreal.  You will also need to bump your version.

src/finiteVolume/processes/les.cpp
Rozie100: I see. Thank!

mmcgurn (approved):

PR #413, opened 2023-01-10
Surface Radiation Unit Testing

owenmylotte:
Compares surface exchange radiation problems with analytical solutions for multiple geometries and MPI process numbers.


owenmylotte:
Requires DMSwarm patch in order to pass tests.

PR #414, opened 2023-01-13
Neural Cpv inversion

profPlum:
Hey this has the latest changes with a fix for inversion issue & constraints to ensure outputs are between 0 & 1 and sum to 1. Also tests are passing & you can use the latest models for wax & methane.


src/chemistry/chemTabModel.cpp
```diff
@@ -0,0 +1,277 @@
+#include "chemTabModel.hpp"
```
mmcgurn, line 1: The chemTabModel was moved and refactored into src/eos/chemTab.hpp with api changes.  You need to move any required changes into this class and delete this one.

mmcgurn (changes requested):
Hi Dwyer.  The chemTabModel was moved a while ago and refactored into src/eos/chemTab.hpp with api changes.  You need to move any required changes into this class.  Your new chemTabTestModel is not being tested.  You need to add it to tests/unitTests/eos/chemTabTests.cpp and remove any `GTEST_SKIP() << "Test will not work until ChemTab can decode progress (with zMix) from Yi";` from the file.  Those tests should now be working.  The chemTabTestModel_1-PMMA is missing the required metadata and would fail the tests.  Can we rename chemTabTestModel_1-PMMA to chemTabTestModel_2?

PR #422, opened 2023-01-20
changed the physics (cfl) based tilmestep to use petsc adapt

mmcgurn:
This is in preparation for a new time step physicsConstrained to being implemented in the future.

Instead of separate setting `ts_adapt_type: none` and `computePhysicsTimeStep: true` just select custom type `ts_adapt_type: physics`


src/finiteVolume/finiteVolumeSolver.cpp
```diff
@@ -252,37 +248,18 @@ void ablate::finiteVolume::FiniteVolumeSolver::RegisterRHSFunction(RHSArbitraryF
 
 void ablate::finiteVolume::FiniteVolumeSolver::RegisterPreRHSFunction(PreRHSFunctionDefinition function, void* context) { preRhsFunctions.emplace_back(function, context); }
 
-void ablate::finiteVolume::FiniteVolumeSolver::EnforceTimeStep(TS ts, ablate::solver::Solver& solver) {
```
owenmylotte, line 255: So each solver gets a ComputeTimeStep function instead of the option that existed before?

src/finiteVolume/processes/navierStokesTransport.cpp
```diff
@@ -218,7 +218,7 @@ double ablate::finiteVolume::processes::NavierStokesTransport::ComputeCflTimeSte
 
     // Get the valid cell range over this region
     solver::Range cellRange;
-    flow.GetCellRange(cellRange);
+    flow.GetCellRangeWithoutGhost(cellRange);
```
owenmylotte, line 221: This is definitely worth having!

owenmylotte (approved):
This definitely looks cleaner!

Rozie100 (approved):
Makes sense!

src/finiteVolume/finiteVolumeSolver.cpp
mmcgurn: Yes, every solver get the option to implement a physics based timeStep

PR #430, opened 2023-02-08
Calculates two phase CFL

maesemen:
Allows twoPhaseAdvection solver use physics time step.  Also fixes typo in eos::TwoPhase.

### DIFF
--- a/src/finiteVolume/processes/twoPhaseEulerAdvection.cpp
+++ b/src/finiteVolume/processes/twoPhaseEulerAdvection.cpp
@@ -3,6 +3,7 @@
 #include <utility>
 #include "eos/perfectGas.hpp"
 #include "eos/stiffenedGas.hpp"
+#include "eos/twoPhase.hpp"
 #include "parameters/emptyParameters.hpp"
 #include "finiteVolume/compressibleFlowFields.hpp"
 #include "flowProcess.hpp"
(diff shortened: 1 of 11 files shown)


src/finiteVolume/processes/twoPhaseEulerAdvection.cpp
```diff
@@ -191,6 +193,7 @@ PetscErrorCode ablate::finiteVolume::processes::TwoPhaseEulerAdvection::FormJaco
 }
 
 ablate::finiteVolume::processes::TwoPhaseEulerAdvection::TwoPhaseEulerAdvection(std::shared_ptr<eos::EOS> eosGas, std::shared_ptr<eos::EOS> eosLiquid,
```
mmcgurn, line 195: I would pass the two phase equation of state into this instance instead of the separate separate equation of states.  This way the same equation of state can be used in other functions/classes

src/finiteVolume/processes/twoPhaseEulerAdvection.cpp
```diff
@@ -294,6 +307,68 @@ PetscErrorCode ablate::finiteVolume::processes::TwoPhaseEulerAdvection::Multipha
...
+
+    // Get the dim from the dm
+    PetscInt dim;
+    DMGetDimension(dm, &dim) >> utilities::PetscUtilities::checkError;
+
+    // assume the smallest cell is the limiting factor for now
+    const PetscReal dx = 2.0 * minCellRadius;
+
+    // Get field location for euler and densityYi
+    auto eulerId = flow.GetSubDomain().GetField("euler").id;
```
mmcgurn, line 339: Can you replace this with the variable instead of a raw string?

src/eos/twoPhase.cpp
```diff
@@ -26,14 +26,14 @@ static inline void SimpleGasGasDecode(PetscInt dim, ablate::eos::TwoPhase::Decod
         rho2 = rho;
         e2 = e;
         p = (gamma2 - 1) * rho2 * e2;
-        T = e2 * cv2;
+        T = e2 / cv2;
```
mmcgurn, line 29: Were these wrong?  Can you add some tests that would have caught this bug?

mmcgurn (changes requested):
Thanks for the PR, looks good!  I have a couple of small changes if you don't mind.

src/eos/twoPhase.cpp
maesemen: yes, I can add some test cases for this

mmcgurn (approved):

PR #431, opened 2023-02-08
two phase field functions for Euler and densityvolumeFraction fields

maesemen:


src/eos/twoPhase.hpp
```diff
@@ -16,6 +16,7 @@ class TwoPhase : public EOS {  // , public std::enabled_shared_from_this<TwoPhas
     const std::shared_ptr<eos::EOS> eos2;
     // this mixed eos does not allow species, get species from eos1 eos2
     std::vector<std::string> species;
+    std::vector<std::string> otherPropertiesList = {"VF"};  // otherProperties must include volumeFraction for Field Function initialization
```
mmcgurn, line 19: Is this string defined somewhere else?  Can we use a global variable string instead?

mmcgurn (approved):
Looks good, I have one small request.

PR #433, opened 2023-02-09
Virtual TCP and Arbitrary Surface Radiation Monitor

owenmylotte:
Allows the creation of windows or virtual plates to calculate the radiation that would be detected by a plate or camera. Multiple ray tracing inputs can be input to the same monitor to allow for multiple outputs of different ray tracing behaviors or absorption models.


owenmylotte:
I finished this for now and added the additional check. I still need to make sure that the integration test is in and comparing properly.

src/monitors/radiationFlux.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    // update the name
+    name = radiationFluxRegion->GetName() + name;
+
+    DMLabel radiationFluxRegionLabel = nullptr;
+    PetscInt regionValue = 0;
+    domain::Region::GetLabel(radiationFluxRegion, solverIn->GetSubDomain().GetDM(), radiationFluxRegionLabel, regionValue);
+
+    // Now create a sub dm with only the faces
+    DMPlexFilter(solverIn->GetSubDomain().GetDM(), radiationFluxRegionLabel, 1, &fluxDm) >> utilities::PetscUtilities::checkError;
```
mmcgurn, line 19: Instead of using one here, you can use the region value you pulled on line 16.  Most often it is 1, but not always.

src/monitors/radiationFlux.cpp
```diff
@@ -0,0 +1,181 @@
...
+        DMPlexGetHeightStratum(fluxDm, 0, &cStart, &cEnd) >> utilities::PetscUtilities::checkError;
+
+        // get the mapping information
+        IS faceIs;
+        const PetscInt* faceToBoundary = nullptr;
+        DMPlexGetSubpointIS(fluxDm, &faceIs) >> utilities::PetscUtilities::checkError;
+        ISGetIndices(faceIs, &faceToBoundary) >> utilities::PetscUtilities::checkError;
+
+        for (PetscInt c = cStart; c < cEnd; ++c) {
+            for (int i = 0; i < int(radiation.size()); i++) {
```
mmcgurn, line 143: I don't see any reason to cast to an int here, I would std::size_t instead of int.

tests/integrationTests/inputs/monitors/radiationFlux.yaml
```diff
@@ -0,0 +1,105 @@
+---
```
mmcgurn, line 1: this input looks like one that ablate had already processed?  Do you the the original that used text based anchors?

mmcgurn (changes requested):
Thanks.  This looks great!  I have two small nick-picky things.

mmcgurn (approved):

PR #439, opened 2023-02-23
fixed bug in in-region domain check

mmcgurn:


owenmylotte (approved):
Would the size output garbage if the IS function never writes to the uninitialized variable? That makes a lot of sense.

PR #451, opened 2023-03-15
slab burner 3D regression test

mmcgurn:


src/finiteVolume/processes/navierStokesTransport.cpp
```diff
@@ -362,8 +375,11 @@ double ablate::finiteVolume::processes::NavierStokesTransport::ComputeViscousDif
     auto diffusionData = (DiffusionTimeStepData*)ctx;
 
     // Get the fv geom
-    PetscReal minCellRadius;
-    DMPlexGetGeometryFVM(dm, NULL, NULL, &minCellRadius) >> utilities::PetscUtilities::checkError;
```
kolosret, line 366: Good idea!

tests/regressionTests/inputs/slabBurner2D/slabBurner2DMeshGen.py
```diff
@@ -75,38 +76,60 @@ def defineBoundary(sides, name, boundary_list):
...
 # set the options to prevent gmsh from adding too many elements to each geometry line
 gmsh.option.setNumber("Mesh.MeshSizeFromPoints", 0)
 gmsh.option.setNumber("Mesh.MeshSizeFromCurvature", 0)
 gmsh.option.setNumber("Mesh.MeshSizeExtendFromBoundary", 0)
+gmsh.option.setNumber("Mesh.Smoothing", 10)
 
 # generate the mesh
 gmsh.model.mesh.setRecombine(2, surface_id)
 gmsh.model.mesh.generate(2)
-gmsh.model.mesh.reverse() # we have to reverse the element orientation for petsc
```
kolosret, line 96: Do you have to sync again here?

mmcgurn: I think that gmsh.model.geo.synchronize() is only used for the cad.  https://gitlab.onelab.info/gmsh/gmsh/blob/master/tutorials/python/t1.py States that ```Before they can be meshed (and, more generally, before they can be used by API functions outside of the built-in CAD kernel functions)```

kolosret: Okay great I thought you have to use them every time the gmsh object is called.  Thank you.

kolosret (approved):
Looks good overall!

PR #452, opened 2023-03-22
Salac/get range

dsalac:
Moved all range calls to a central location.


mmcgurn (changes requested):
Thanks for making these changes. Do we know if anyone calls that directly or should those just live in the subdomain?

mmcgurn (approved):

PR #456, opened 2023-04-07
new soot version

mmcgurn:


src/finiteVolume/conservedField.hpp
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Helper class to create a conserved and aux field
+ */
+class ConservedField : public domain::FieldDescriptor {
+   protected:
+    const std::string name;
+    const std::vector<std::string> components;
+    const std::shared_ptr<domain::Region> region;
+    const std::shared_ptr<parameters::Parameters> conservedFieldOptions;
+    const std::shared_ptr<parameters::Parameters> auxFieldOptions = ablate::parameters::MapParameters::Create({{"petscfv_type", "leastsquares"}, {"petsclimiter_type", "none"}});
```
owenmylotte, line 22: Are aux fields always set to this type?

src/finiteVolume/processes/soot.cpp
```diff
@@ -0,0 +1,397 @@
...
+        const PetscScalar* conserved = nullptr;
+        PetscCall(DMPlexPointGlobalRead(flow.GetSubDomain().GetDM(), cell, solutionArray, &conserved));
+
+        // If a real cell (not ghost)
+        if (conserved) {
+            // store the data for the chemistry ts (T, Yi...)
+            PetscReal* temperature;
+            PetscCall(DMPlexPointLocalFieldRead(temperatureDm, cell, temperatureField.id, temperatureArray, &temperature));
+
+            if (*temperature > soot->thresholdTemperature) {
```
owenmylotte, line 186: If we're doing things like this is it worth considering dynamic load balancing of the mesh? I wonder how hard that would be to implement.

src/finiteVolume/processes/soot.cpp
```diff
@@ -0,0 +1,397 @@
...
+    // Add in the Soot Reaction Sources
+    real_type SVF = xArray[C_s] * pointInfo->currentDensity / solidCarbonDensity;
+
+    // Total S.A. of soot / unit volume
+    PetscReal SA_V = calculateSurfaceArea_V(xArray[C_s], xArray[ODE_NDD], pointInfo->currentDensity);
+
+    // Need the Concentrations of C2H2, O2, O, and OH
+    // It is unclear in the formulations of the Reaction Rates whether to use to concentration in regards to the total mixture or just the gas phace, There is a difference due to the density relation
+    //-> For now we will use the concentration to be the concentration in the gas phase as it makes more physical sense
+    PetscReal C2H2Conc = pointInfo->currentDensity * PetscMax(0, xArray[C2H2]) / pointInfo->mw[C2H2];
```
owenmylotte, line 337: Have you considered making these functions and the stuff after them children of an interface that can be iterated over? Just so it might look cleaner.

tests/regressionTests/inputs/slabBurner2D/slabBurner2D.yaml
```diff
@@ -85,15 +85,20 @@ timestepper:
       # class that creates the required fields for the compressible flow solver (rho, rhoE, rhoU, ...)
       - !ablate::finiteVolume::CompressibleFlowFields
         # The soot equation of state is an extension of the tChem eos with support for a simple soot model
-        eos: !ablate::eos::TChemSoot  &eos
-          mechFile: MMAReduced.yml
+        eos: !ablate::eos::TChem  &eos
```
owenmylotte, line 88: Is there no more soot eos?

tests/unitTests/eos/radiationProperties/radiationSootAbsorptionTests.cpp
```diff
@@ -40,23 +42,25 @@ TEST_P(SootTestFixture, ShouldProduceExpectedValuesForField) {
...
+INSTANTIATE_TEST_SUITE_P(RadationSootTests, SootTestFixture,
+                         testing::Values(/** A test with all valid species for the Soot model */
+                                         (SootTestParameters){.fields = {ablateTesting::domain::MockField::Create("euler", 5, 0),
+                                                                         ablateTesting::domain::MockField::Create("densityYi", {ablate::eos::TChemSoot::CSolidName}, 5)},
+                                                              .conservedValues = {0.01, NAN, NAN, NAN, NAN, 0.0025},  //!< The Density Yi values live here
+                                                              .temperatureIn = 300.0,
+                                                              .densityIn = 0.01,  //!< The density is read through the equation of state above, not here
+                                                              .expectedAbsorptivity = 0.67868822124163297},
+                                         /** A test with three valid species for the Soot model. */
+                                         (SootTestParameters){.fields = {ablateTesting::domain::MockField::Create("euler", 5, 0),
```
owenmylotte, line 54: Thanks

owenmylotte (approved):

tests/regressionTests/inputs/slabBurner2D/slabBurner2D.yaml
mmcgurn: There is, but it is not needed when using the new soot process.

src/finiteVolume/conservedField.hpp
mmcgurn: Yes, but default they are.

src/finiteVolume/processes/soot.cpp
mmcgurn: Yeah I think that is a good idea.  This was a direct port of existing code, if we are happy with the results we should go back and clean this up.

PR #471, opened 2023-06-06
Afashami/rbf vertex

afashami:
Update rbfSupport.cpp and modify the external function in order to get either neighbor vertices or cells.


dsalac:
Please merge the latest version of UBCHREST:main into afashami/rbf_vertex to include PETSc API changes.

PR #480, opened 2023-07-11
marziale/rbfSupport to petscSupport

jjmarzia:
In the current version, the rbfSupport file contains functions that are not related to actual radial basis functions or RBFs. So in this updated version, these functions along with their respective unit test files are moved into a more general petscSupport, to be eventually merged in with PETSc.


mmcgurn (approved):
Thanks!

src/utilities/petscSupport.cpp
```diff
@@ -0,0 +1,1093 @@
...
+            N[0] = PetscSignReal(edgeCenter[0] - vCoords[0]);  // In 1D this is simply the vector connecting the vertex to the edge center
+            break;
+        case 2:
+            DMPlexSurfaceAreaNormal2D_Internal(dm, vCoords, edgeCenter, nFace, faces, N);
+            break;
+        case 3:
+            DMPlexSurfaceAreaNormal3D_Internal(dm, -1, vCoords, edgeCenter, nFace, faces, N);
+            break;
+        default:
+            SETERRQ(PetscObjectComm((PetscObject)dm), PETSC_ERR_ARG_OUTOFRANGE, "DMPlexEdgeSurfaceAreaNormal can not handle dimensions of %d", (int)dim);
```
mmcgurn, line 708: You can use the PetscInt_FMT to handle formatting petsc ints when int or long.  This is an example in the code https://github.com/UBCHREST/ablate/blob/350c3a540fd75b7aa33c87e7818a22ce889398f2/src/solver/solver.cpp#L105

PR #489, opened 2023-08-02
Salac/riemann refactor clean

dsalac:
Reduced common code to a single calls that are used across different Riemann solvers.


src/finiteVolume/fluxCalculator/riemannCommon.hpp
```diff
@@ -0,0 +1,27 @@
...
+ * @param aR: speed of sound on the right center cell
+ * @param rhoR: density on the right cell center
+ * @param p0R: reference pressure for stiffened gas on right (pass in from EOS). Will be zero if perfect gas or single gas
+ * @param pR: pressure on the right cell center
+ * @param gammaR: specific heat ratio for gas on right (pass in from EOS)
+ * @param pstar0: initial guess at the pressure across contact surface
+ * @param massFlux: mass flux across the cell face(?)
+ * @param p12: interface pressure(?)
+ */
+ablate::finiteVolume::fluxCalculator::Direction riemannSolver(const PetscReal uL, const PetscReal aL, const PetscReal rhoL, const PetscReal p0L, const PetscReal pL, const PetscReal gammaL,
```
mmcgurn, line 23: For something like we I think we might want to keep riemannSolver function out of the global name space.  Would you mind if we moved it to a static function inside of Riemann?

mmcgurn (changes requested):
Thanks for cleaning this up.  I have just one request for function location.

src/finiteVolume/processes/surfaceForce.cpp
```diff
@@ -4,6 +4,11 @@
 #include "utilities/constants.hpp"
 #include "utilities/mathUtilities.hpp"
 
+#define xexit(S, ...)                                                                                                                                                 \
```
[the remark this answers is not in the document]
dsalac: Dammit. Forgot to remove that.

src/finiteVolume/fluxCalculator/riemannCommon.hpp
dsalac: This is just to get access to the direction enum. The file riemann.cpp contains the single ideal gas. Is this where you were thinking of moving it to?

dsalac: Sure. I can make the Riemann solver a sub-class of fluxCalculator and the three cases sub-classes of the Riemann solver.

maesemen (approved):

mmcgurn (approved):
Thanks

PR #494, opened 2023-08-23
Salac/vof level set yaml

dsalac:
I'm sure there is a better way to pull the output file in vofFromLS_yaml.cpp: 70


mmcgurn:
> I'm sure there is a better way to pull the output file in vofFromLS_yaml.cpp: 70

I'm not sure if you need to read it it in, the solution vector should be memory.  I could add a testing_monitor that computes the difference at the end of the simulation so we could move this the normal integration tests.  Please let me know if you would like me to do this.

dsalac:
That's OK with me. Whatever makes the most sense.

PR #500, opened 2023-10-30
fix that prevents from building on Mac.

mmcgurn:
This is part of the solution, check wiki for change in dependencies.


jikembo (approved):
So HDF5 was the only thing missing.... Thank you for catching it.

mmcgurn:
@jikembo Would you mind trying to build with the latest directions on the wiki?

PR #509, opened 2023-12-14
Surface absorptivity addition

klbudzin:
Added the option to specify a surface absorptivity which defaults to the emissivity if not specified. A critical impact of this change is that the surface radiation emissivities now have to be specified as an input parameter and will need to be altered on all pre-existing input files using surface emission, or else it will default to 1!

### DIFF
--- a/src/boundarySolver/physics/sublimation.cpp
+++ b/src/boundarySolver/physics/sublimation.cpp
@@ -24,9 +24,8 @@ ablate::boundarySolver::physics::Sublimation::Sublimation(std::shared_ptr<subMod
       pressureGradientScaling(std::move(pressureGradientScaling)),
       radiation(std::move(radiationIn)),
       radiationInterval((intervalIn ? intervalIn : std::make_shared<io::interval::FixedInterval>())),
-      emissivity(parametersIn ? parametersIn->Get<PetscReal>("emissivity",1.) : 1.),
-      absorptivity(parametersIn ? ( parametersIn->Get<PetscReal>("absorptivity",-1) == -1 ?
-              parametersIn->Get<PetscReal>("emissivity",1.) : parametersIn->Get<PetscReal>("absorptivity",-1.) ) : 1.),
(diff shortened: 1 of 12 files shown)


src/boundarySolver/physics/sublimation.cpp
```diff
@@ -23,7 +23,8 @@ ablate::boundarySolver::physics::Sublimation::Sublimation(std::shared_ptr<subMod
       pressureGradientScaling(std::move(pressureGradientScaling)),
       radiation(std::move(radiationIn)),
       radiationInterval((intervalIn ? intervalIn : std::make_shared<io::interval::FixedInterval>())),
-      emissivity(emissivityIn == 0 ? 1.0 : emissivityIn),
+      emissivity(parametersIn ? parametersIn->Get<PetscReal>("emissivity", 1.) : 1.),
+      absorptivity(parametersIn ? (parametersIn->Get<PetscReal>("absorptivity", -1) == -1 ? parametersIn->Get<PetscReal>("emissivity", 1.) : parametersIn->Get<PetscReal>("absorptivity", -1.)) : 1.),
```
mmcgurn, line 27: Instead of doing the -1 double comparisons we may want to grab GetString from the parameters.  It returns an optional if available, so you can treat it as the left hand side of the conditional.

mmcgurn (changes requested):
Looks great! Thanks!  I just have one comment on reading from the parameters.  You will also need to bump the ABLATE version.

src/boundarySolver/physics/sublimation.cpp
klbudzin: That makes sense, that'll make it a bit more readable!  I'll make the change and get back to you. I'm also going to add in a 1DDiffusionMonitor that can be used with 1D diffusion flames for coupled boundaries that will just return the mass flux to a log gable if that makes sense to do.

mmcgurn (approved):
Thanks for fixing these things!

PR #513, opened 2024-01-22
axisymmetric cylinder mesh

mmcgurn:
This PR adds support for arbitrary  axisymmetric cylinder meshes.  Also support for other meshes that supply cell/node values.  @klbudzin Would you mind taking a look at this PR and seeing if the results for axisymmetricPipeFlow look reasonable? You will have to extend the max time steps if you run it locally.


klbudzin (commented):
Everything Here looks good to me now!

PR #514, opened 2024-02-06
Updated expected cells for a few tests.

dsalac:
What's interesting is that the test didn't pick up the change all the time. I could run it 4 times in a row showing it passed and then it would fail the next 5 times.


mmcgurn:
I’ll force the petsc docker image to rebuild so we run the tests against the latest.

dsalac:
Thanks. What version of clang are we using for the format-check? It's throwing a bunch of errors for me in other files.

mmcgurn:
I think the tests are still using clang 16 with older .clang-format.v16 file in the config folder.  The configuration selects which config file based upon the version.  I use clang 17 with the default format file on my machine.  So far the results have been the same.

mmcgurn:
Merged this code as part of PR #516

PR #515, opened 2024-02-06
Add support for source terms to cells from particles.

mmcgurn:
@klbudzin This PR adds support for particle source terms to the flow/cell field.  It only works with fields with one component but this should be fixed by PETSc in the future. It requires the petsc branch knepley/fix-swarm-fv for now.


mmcgurn:
@klbudzin The particle push should now work but other tests are broken.  I think we are going to remove a special case/hack in PETSc because it is no longer needed but you should be able to start developing based upon this branch.  If you don't mind take a look at the test with the push and see if the value conserves as you would expect.

PR #517, opened 2024-02-22
Salac/dmplexupdates

dsalac:
Previous version had issues with ghost cells.


dsalac:
Going to clean up commit history first.

PR #518, opened 2024-02-22
Updated some petsc support functions to work with ghost cells

dsalac:
Updated some functions to support ghost cells. Kept DMPlexCellGetNumVertices simply for compatibility.


mmcgurn (approved):
Thanks!  I thinking there are some formatting issues and you need to bump the version.  If you would like I can do those things and PR back into your branch.

dsalac:
Just updated the formatting. I was using clang-19, which doesn't look compatible with the tests.

mmcgurn:
Thanks! We might need to add a new clang-format file to maintain backwards compatibility.  I did this with version 16, but we might want to leave the default clang-format file to be compatible with clion.

PR #519, opened 2024-02-29
Generalized Axis specification for AxiSymmetric Mesh

mmcgurn:
@klbudzin I think the multiple meshes we were talking about was getting to convoluted.  I'm thinking we might want to get this in and explore generating a AxiSymmetric mesh from gMesh.  I can work on this week/beginning of next if you agree.


klbudzin (approved):
@mmcgurn Everything here looks great! Using these mappers we should be able to at least cluster the elements closer to the fuel in the fuel region for now. I do think it's worthwhile to just get the gmsh axisymmetric parser in there as soon as possible like you said as well because the logic for doing the vertical slice constant radius meshing seems very non-intuitive, and being able to have less shells in the valve/tubing region would be nice to not have our minimum global step size be extremely squished there killing time stepping when it isn't really needed.

klbudzin:
Also I do have a form of the outer boundary shell function for different regions on a branch on my fork I've been using. Did you want me to rebase it and create a merge request or wait and have it combined with a future one.

mmcgurn:
I agree with your thoughts on the gMsh reader.  Yes, if you don't mind rebasing your code on top of this I would appreciate it.

PR #525, opened 2024-03-28
Adding Support for defining different boundary region on outer shell boundary

klbudzin:
Added in support for a user defined math function that can be used to label different boundary values on the outershell of an axisymmetric mesh based on the Z coordinate.

### DIFF
--- a/src/domain/descriptions/axisymmetric.cpp
+++ b/src/domain/descriptions/axisymmetric.cpp
@@ -121,8 +121,8 @@ std::shared_ptr<ablate::domain::Region> ablate::domain::descriptions::Axisymmetr
 
     // compute the outer shell start
     auto outerShellStart = numberCenterVertices + numberVerticesPerShell * (numberShells - 1);
-    //Coordinate used to track the z location
-    auto* coordinate = new PetscReal[]{0.0,0.0,0.0};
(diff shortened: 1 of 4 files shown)


kolosret (commented):
Makes sense to me.

src/domain/descriptions/axisymmetric.cpp
```diff
@@ -120,7 +121,9 @@ std::shared_ptr<ablate::domain::Region> ablate::domain::descriptions::Axisymmetr
 
     // compute the outer shell start
     auto outerShellStart = numberCenterVertices + numberVerticesPerShell * (numberShells - 1);
-
+    // Coordinate used to track the z location
+    auto *coordinate = new PetscReal[]{0.0, 0.0, 0.0};
```
mmcgurn, line 125: This is being deleted anywhere?  Can we create it only on the stack?

mmcgurn (changes requested):
One small question, otherwise looks great!

src/domain/descriptions/axisymmetric.cpp
klbudzin: Good catch, I didn't realize I initialized it that way... It wasn't being deleted anywhere, but I fixed it to be created on the stack.

PR #527, opened 2024-07-31
Soot and petsc fixes

klbudzin:
Recent PETSc update added deprecated petsc calls that needed fixing. Also added in the ability to change some rates on the soot mechanism during run time for optimization/testing. Also included is an addition to run the decode steps before the first the initialization output for quality of life when looking at the initialized fields.


src/finiteVolume/processes/soot.hpp
```diff
@@ -13,6 +13,9 @@ class Soot : public Process {
     [[maybe_unused]] inline const static double solidCarbonDensity = 2000;
     //! Scaling term for Ndd going into the Tines ODE Solver
     [[maybe_unused]] inline static double NddScaling = 1e20;
+    //! Parameters for Nucleation and Surface Growth Pre Exponentials
+    [[maybe_unused]] inline static double nucPreExponential = 1000;
```
kolosret, line 17: arent these always used?

kolosret (approved):
Looks good!

klbudzin:
I decided to just turn off the 5 compressible flow tests that are failing using PETSc's periodic boundary conditions and to open up an issue to fix them after I figure out what's holding back our highly discretized mesh's from initializing properly. This way anyone can still develop using the newest petsc builds.

PR #538, opened 2025-05-01
Adding user defined limiter fix, MF boundary and profiling improvements

kolosret:
1, Added maximum gradient option in cellinterpolant, over which we don't project the state to the face, we use cell-centered values. Recommended value is around 10000 for user.
2. Added mass flux boundary condition form Kenny's branch.
3. Added barrier in the timestepper between pre and evaluate step. This is to stop radiation gain load imbalance influencing flow data.

### DIFF
--- a/src/finiteVolume/cellInterpolant.cpp
+++ b/src/finiteVolume/cellInterpolant.cpp
@@ -2,7 +2,8 @@
 #include <petsc/private/dmpleximpl.h>
 #include <utility>
 
-ablate::finiteVolume::CellInterpolant::CellInterpolant(std::shared_ptr<ablate::domain::SubDomain> subDomainIn, const std::shared_ptr<domain::Region>& solverRegion, Vec faceGeomVec, Vec cellGeomVec,double maxGradIn)
+ablate::finiteVolume::CellInterpolant::CellInterpolant(std::shared_ptr<ablate::domain::SubDomain> subDomainIn, const std::shared_ptr<domain::Region>& solverRegion, Vec faceGeomVec, Vec cellGeomVec,
+                                                       double maxGradIn)
     : subDomain(std::move(std::move(subDomainIn))), maxLimGrad(maxGradIn) {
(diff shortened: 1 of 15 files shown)


src/finiteVolume/finiteVolumeSolver.hpp
klbudzin, line 1: If we are setting in the finite volume solver, it should have a defaulted value here, which should be some ridiculously large value that should probably be linked to the ablate::utilities::constants::large value

src/finiteVolume/cellInterpolant.cpp
```diff
@@ -450,11 +454,25 @@ void ablate::finiteVolume::CellInterpolant::ComputeFieldGradients(const domain::
                 DMPlexApplyLimiter_Internal(dm, dmCell, lim, dim, dof, cell, field.id, cellFaces[f], faceRange.start, faceRange.end, cellPhi, xLocalArray, cellGeometryArray, cg, cx, cgrad) >>
                     utilities::PetscUtilities::checkError;
             }
+
             /* Apply limiter to gradient */
+            PetscBool cancel = PETSC_FALSE;
+            for (PetscInt pd = 0; pd < dof; ++pd) {
+                if (cellPhi[pd] == 0) {
+                    for (PetscInt d = 0; d < dim; d++) {
+                        // limit the maxium gradient, this is required for strong shocks in rocket simulations
```
klbudzin, line 463: Probably expand on this comment, i.e. change it to something similar to
 // Due to directional limiting being difficult in unstructured grids, a strict gradient limiter is introduced here to revert back to cell centered reconstructions if a component limiter is 0 even though there is a strong component gradient in a direction (Usually happens at strong shocks seen in high pressured rocket simulations)

klbudzin (changes requested):

src/finiteVolume/finiteVolumeSolver.hpp
kolosret: sounds good, changed it!

klbudzin (approved):

